PR #1, opened 2023-04-17
feat: generated protobuf classes

danepitkin:
A python package containing the generated protobuf classes.

* Use `buf` and `protoletariat` to generate the protobuf classes
* Use setuptools for python packaging and dynamic versioning
* Pin Substrait submodule to latest release (v0.28.2)

Testing:
* Ran through README instructions to generate proto, build python package, run pytests, and import module locally.


tests/test_proto.py
```diff
@@ -0,0 +1,10 @@
+def test_imports():
+    from pysubstrait.proto.pysubstrait.algebra_pb2 import Expression
```
vibhatha, line 2: Just curious about the `pysubstrait.proto.pysubstrait...` could we make it something like
`from pysubstrait.proto.core.algebra_pb2 import Expression` or just `from pysubstrait.proto.algebra_pb2 import Expression`

README.md
```diff
@@ -0,0 +1,207 @@
...
+## Python package
+### Editable installation
+```
+pip install -e .
+```
+
+## Generate protocol buffers
+Generate the protobuf files manually. Requires protobuf `v3.20.1`.
+```
+./gen_proto.sh
```
vibhatha, line 187: Just curious: 

Could we make this part of the setup which would automatically run?

gforsyth: I don't think we want to do that -- one of the conveniences here should be that the files are already generated (using the correct version of `protobuf`) and bundled into the package for the end-user.

tests/test_proto.py
danepitkin: I would also have preferred that, but the second `pysubstrait` was added as a namespace for the generated protobuf classes so as not to conflict with other implementations like C++. I used this script in `gen_proto.sh` to generate it: https://github.com/substrait-io/substrait/blob/main/tools/proto_prefix.py. I'm open to a better naming convention if anyone has one!

vibhatha: Aha, I understand your point. :thinking:

danepitkin: We can always merge as-is and change it in the future while the project is still pre- v1.0. For reference, Ibis-substrait sort of does it this way, too: https://github.com/ibis-project/ibis-substrait/tree/main/ibis_substrait/proto/substrait/ibis

danepitkin:
> I think this is a good starting point and the general structure is good in re: protobuf generation, namespacing, etc.
> 
> My preference would be to call the module and the pypi package `substrait` instead of `pysubstrait`, but that's not a blocker.

Since there are a lot of opinions on the two names, is it worth voting in the ML? It's probably best that we definitively pick one early on.

README.md
westonpace, line 187: Are these instructions for `pysubstrait` developers?  Or are these instructions for `pysubstrait` users?

If these are for developers should they be in a `CONTRIBUTING.md`?  If they are for users then why does it list `gen_proto.sh`?  Is that something a user has to do?

tests/test_proto.py
westonpace: Actually "standard" (as in, the message defining the hint is in the core substrait proto files) hints will probably be more of a danger than extensions.

westonpace (approved):
Looks like a fine starting point to me.

tests/test_proto.py
gforsyth: I think we haven't actually run into this in ibis-substrait yet, but I don't know that anyone has tried it.

README.md
danepitkin: Everything at and below "Getting started" is for devs! I'll move it to `CONTRIBUTING.MD`. Great idea!

tests/test_proto.py
danepitkin: Do you mind elaborating on the dangers with an example?

westonpace: Sure.  Let's pretend I created a message (`EstimatedSelectivity`) that was intended to be used as an advanced extension on `RelCommon::Hint` and we merged it into Substrait:

```
message RelCommon {
  ...
  Hint hint = 3;
  ...
  message Hint {
    ...
    substrait.extensions.AdvancedExtension advanced_extension = 10;
    ...
  }
  ...
}

message EstimatedSelectivity {
  double selectivity = 1;
}
```

Then, I attached that hint onto a filter relation.  Here is the plan, serialized as JSON:

```
{
  ...
  "filter": {
    "common": {
      "hint": {
        "optimization": {
          "@type": "substrait.EstimatedSelectivity",
          "selectivity": 0.03
        }
      }
    },
    "condition": { ... },
    "input": { ... }
  }
  ...
}
```

As you can see, the Substrait package is included as part of the `@type` when an `Any` is serialized.

However, I don't know of any existing use like this.  In Acero we use `google.protobuf.Any` to add new nodes (currently as-of join and segmented aggregation).  In both cases the new message is not part of the core Substrait set of messages.  It's a part of Acero's custom protobuf.  This custom protobuf uses Substrait messages (e.g. an as-of join has field references in it and these are Substrait expressions) but the only type name that gets serialized is the extension itself.

I think it's reasonable to expect that will never be the case.  For example, consider the above example.  One could argue that a new "standard hint" should be included in the hint definition as:

```
message RelCommon {
  ...
  Hint hint = 3;
  ...
  message Hint {
    ...
    EstimatedSelectivity estimated_selectivity = 11;
    substrait.extensions.AdvancedExtension advanced_extension = 10;
    ...
  }
  ...
}
```

In other words, any time `google.protobuf.Any` is used it must refer to an object that is not part of the core Substrait protobuf files.

However, I'm not completely convinced it will never happen.

drin: I believe what Weston is describing is how I planned to approach doing my own extensions. I think the most concise example is here, where I use ibis-substrait to translate and then I use a `google.protobuf.Any` for an `ExtensionLeafRel`:
https://github.com/drin/mohair-extension/blob/mainline/mohair_extension/extension.py#L100

Using just the standard string representation, here's what I generate:
https://media.githubusercontent.com/media/drin/mohair-extension/mainline/resources/projection.mohair.txt


> an object that is not part of the core Substrait protobuf files

I am assuming that if something becomes core, then I would change the way this is done. Using ibis-substrait's approach to translation makes it an easy change, not sure that approach holds true for any other, large projects.

drin (approved):
looks good to me.

tests/test_proto.py
danepitkin: Thank you, both! The examples are super helpful. It's a good thing to note!

vibhatha:
Looks good to me.

vibhatha (approved):

gforsyth:
> Since there are a lot of opinions on the two names, is it worth voting in the ML? It's probably best that we definitively pick one early on.

I haven't familiarized myself with voting procedures for the project, but it's an option.  

I think it's worth looking at what other languages have done:

Java (original implementation), refers to itself as `io.substrait`: https://mvnrepository.com/artifact/io.substrait
Rust, refers to itself as `substrait`: https://docs.rs/substrait/latest/substrait/
Go, refers to itself as `substrait-go`: https://pkg.go.dev/github.com/substrait-io/substrait-go 
C++, uses the namespace `io::substrait::common`: https://github.com/substrait-io/substrait-cpp/blob/main/include/substrait/common/Exceptions.h#L9

All but `go` elide the language abbreviation from the name, and I believe `go` automatically removes `go-` and `-go` when referring to libraries internally (needs a fact check).

westonpace:
>  C++, uses the namespace io::substrait::common: https://github.com/substrait-io/substrait-cpp/blob/main/include/substrait/common/Exceptions.h#L9

To be fair, the cmake project name (which is probably a closer equivalent) is actually substrait-cpp.  That being said, I'm in favor of `substrait` over `pysubstrait`.

gforsyth:
> To be fair, the cmake project name (which is probably a closer equivalent) is actually substrait-cpp.

Sorry!  I didn't mean to misrepresent :sweat_smile:

environment.yml
```diff
@@ -0,0 +1,12 @@
+name: pysubstrait
```
danepitkin, line 1: How do we feel about leaving the conda environment as pysubstrait? Alternatively we can rename to the repo: `substrait-python`

tests/test_proto.py
```diff
@@ -0,0 +1,10 @@
+def test_imports():
+    from substrait.proto.pysubstrait.algebra_pb2 import Expression
```
danepitkin, line 2: What should the protobuf class namespace be in the package? `pysubstrait`? `python`? `python.substrait`? etc.

drin: i think pysubstrait -> substrait

environment.yml
drin: I have minimal opinion since I don't know what it affects. Naively, it seems most straightforward for it to match the package name

gforsyth: Yeah, this you can leave alone.  A person can always override it if they want to, and it won't impact the package name on `conda-forge` (which should probably be `substrait-python` to follow convention there)

src/substrait/__init__.py
```diff
@@ -1,4 +0,0 @@
-try:
```
danepitkin: Hmm, this gets overwritten by ./gen_proto.sh when we don't use a subdirectory within the package.

src/substrait/__init__.pyi
```diff
@@ -0,0 +1 @@
+from . import proto
```
drin, line 1: I was trying to see if we could move the version info to this `.pyi` file you have, but it seems like pyi files are only supposed to have type information. does it make sense to have this stub file if there's only this import?

src/substrait/__init__.py
drin: hm, I was looking to see if we could move this somewhere, but I'm suddenly unsure why this is getting overwritten if it should be putting files into a `proto` directory.

src/substrait/__init__.pyi
danepitkin: I don't think so. It gets autogenerated from `protol` I believe. It used to generate .pyi files for all the classes, but stopped (maybe my version was updated locally mid-development?). Maybe we just add `*.pyi` to .gitignore.

src/substrait/__init__.py
danepitkin: I think I prefer to keep the generated files in `substrait.proto.<namespace>`

gen_proto.sh
```diff
@@ -2,11 +2,11 @@
...
 set -eou pipefail
 
-namespace=pysubstrait
+namespace=proto
 submodule_dir=./third_party/substrait
 src_dir="$submodule_dir"/proto
-tmp_dir=./proto
-dest_dir=./src/substrait/proto
+tmp_dir=./buf_work_dir
+dest_dir=./src/substrait
```
drin, line 9: I think the hack-y way is to make this an intermediate dir and then rename it or put it in a different namespace (like `pysubstrait`) and then alias the imports via the `__init__.py` file?

drin: JK, the script just calls `protol`. Maybe remove the `--create-package` option and it will stop creating the empty `__init__.py` file

danepitkin: ooh ill give this a try!

danepitkin: It doesn't quite work as we'd like. --create-package will recursively create init files, which we do want (except at the top level). Changing --in-place to --not-in-place would disallow overwriting of files, which we don't quite want either. Ultimately the cleanest approach to me still feels like placing these files in a `proto` subdir.

drin: Unless I'm misunderstanding you, @gforsyth 's suggestion is a `proto` subdir. I just assumed that the `protol` tool is creating an extra `__init__.py` file at the same level as the `proto` subdir.

I imagine the directory layout to be something like:
```
substrait
    ├── __init__.py (protol generates this by default)
    ├── proto (this is the proto subdir)
        ├── algebra_pb2.py
        ├── plan_pb2.py
        ├── ...
```

tests/test_proto.py
```diff
@@ -0,0 +1,11 @@
+def test_imports():
+    """Temporary sanity test"""
+    from substrait.gen.proto.algebra_pb2 import Expression
```
danepitkin, line 3: How do we feel about `substrait.gen.proto`?

gen_proto.sh
drin: but that being said, I see you added a gen directory which is probably fine in the near-term. I'm not sure what's generating that `__init__.py` file

danepitkin: Yep your initial understanding is correct! IMO it would be better to generate files into a subdir if `protol` doesn't support exactly what we want. I'm hesitant to introduce any opaque behavior for devs to reduce bugs in the future. Having to choose between overwriting the top level __init__.py or not recursively generating __init__.py's seems bad. I'd rather have both if it means generating files into a subdir instead of root dir.

danepitkin:
I'm satisfied with this implementation if folks are as well. IMO it is about ready to merge!

gen_proto.sh
danepitkin: Oh and I forgot to mention that `protol` was *deleting* the top level __init__.py file when removing the `--create-package` option!

danepitkin: That was because I had `--in-place` option set (aka overwrite).

vibhatha:
I feel like now it is in a good shape. @gforsyth any thoughts?

jorisvandenbossche (approved):

gforsyth (approved):
:shipit: 

This looks great @danepitkin !  Thanks for accommodating all of the change discussions.

gforsyth:
I'll leave this open for the rest of today-ish in case anyone else wants to take a look, but I'll merge it in tonight.

danepitkin:
Thanks everyone for all the feedback, it was a huge help! Looking forward to seeing how Substrait Python grows from here.

gforsyth:
Thanks for leading the charge on this, @danepitkin !

Issue #3, opened 2023-04-18
Adding CIs


vibhatha:
After the initial PR, we should probably add CIs.

westonpace:
> Do you think https://github.com/substrait-io/substrait-python/pull/5 is enough to close this issue?

Yes

westonpace:
Though maybe lets make a follow-up issue for linters / formatters (which are not part of #5 if I understand)

danepitkin:
Added #6!

PR #5, opened 2023-05-16
feat(cicd): build, test, and release github actions

danepitkin:
* Build the Substrait Python package
* Locally install the package and run pytests
* Version tag triggers github release
* Version tag triggers PyPi publish
* Push to main triggers Test PyPi publish

Requires GH secrets to be added for `PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN`.


gforsyth (commented):
Hey @danepitkin -- this looks good to me overall, and I think the logic is sound.

What do you think about splitting out the release and publish workflows into a separate file?  It's still possible to use `needs` across files and it can avoid certain footguns.

danepitkin:
> Hey @danepitkin -- this looks good to me overall, and I think the logic is sound.
> 
> What do you think about splitting out the release and publish workflows into a separate file? It's still possible to use `needs` across files and it can avoid certain footguns.

I would also prefer this approach, but it started to get a bit hairy w/o myself having a good understanding of how to test these actions. AFAIK `needs` can only be used in the same workflow. When you have separate workflows (files), you need to use `workflow_run` e.g.:

```
name: Release Python package

on:
  workflow_run:
    workflows: [build, test]
    types: [completed]
    branches: [main]
```

I can give it a shot! I tried using https://github.com/nektos/act for testing locally, but ran into issues trying to get the `upload-artifact` action to work locally.

danepitkin:
I could also just switch to a `test.yaml` that runs for PRs and a `release.yaml` that does the build + publish.

jorisvandenbossche:
I would also assume that we would have a separate CI workflow for just the tests (that doesn't necessarily creates wheels first), and which would test on different OS, python versions, possibly versions of dependencies, etc

And in that case, I think keeping a simple test step (essentially to verify that the build artifacts are OK) in the build and release workflow might be OK without splitting it.

jorisvandenbossche:
Yes, so essentially what you also said in your last comment

.github/workflows/build_test_release.yml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          submodules: recursive
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Install package dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ".[test]"
```
jorisvandenbossche, line 52: For this "release" workflow, we might want to explicitly install one of the built packages (in `/dist`)

danepitkin: I updated the new `test.yml` to install from `dist/`. It runs on push as well, but its not gating. Probably good to rely on tests to pass before running the release workflow?

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
+name: Build
```
danepitkin, line 1: The Build workflow will always run on PRs, main branch, and version tags. It will upload the build artifact that can be accessed by downstream workflows with 1 day retention (default is 30 days).

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
+name: Run tests
```
danepitkin, line 1: The Test workflow will run when the Build workflow completes AND is successful. It installs the artifact from the build step and tests on mac/win/linux for py 3.8 through 3.11.

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Release
```
danepitkin, line 1: The Release workflow will run when the Test workflow completes AND is successful AND we are on the main branch. If the git ref is a git tag, create a Github release AND upload package to PyPi. If the git ref is the main branch, upload to Test PyPi.

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  workflow_run:
+    workflows: [build]
+    types: [completed]
```
assignUser, line 6: This is actually a security risk as the `workflow_run` trigger starts a run with elevated permissions, which means that running any kind of untrusted code (e.g. from an outside PR). 

The [default permissions ](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)on the GITHUB_TOKEN are rw on almost everything, so even with no secrets exposed in the env a lot of damage could be done.

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
...
+          submodules: recursive
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+          path: dist/
```
assignUser, line 30: This will not work as the artifact actions can only be used within the same run but `workflow_run` is disjunct from the triggering run. You have to use the api to get the artifact (see [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow))

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+    if: ${{ github.repository == 'substrait-io/substrait-python' }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
```
assignUser, line 22: Would it not make sense to build on the lowest supported version?

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install build dependencies
+        run: |
+          python -m pip install build --user
+      - name: Build package
+        run: |
+          python -m build
```
assignUser, line 28: IIUC this is a pure python project at the moment and will build a `none-any` wheel? Otherwise cibuildwheel would be my goto :D

assignUser (changes requested):
Very nice to see such an automatic release process :D 

One major security issue on the test workflow and some additional comments.

The only way to really test these is to merge them to main on your fork and test them there (the pypi steps would fail but as you are relying on an action there it should be fine to assume they will work given the right creds)

.github/workflows/build.yml
jorisvandenbossche: For our use case, that shouldn't matter

.github/workflows/test.yml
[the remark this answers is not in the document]
jorisvandenbossche: Or just have this test workflow not rely on the build workflow. I think we can install the package from source here in one of the first steps (that essentially do that same, but without the need to upload/download artifacts, and this is a very fast step anyway for a small pure python package like we have)

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Install wheel and test dependencies
+        run: |
+          python -m pip install --upgrade pip
+          find ./dist/*.whl | xargs python -m pip install
```
jorisvandenbossche, line 34:
```suggestion
          python -m pip install .
```

Install from source (which will also go through building a wheel, but just on the fly without having to download the pre-built wheel)

.github/workflows/test.yml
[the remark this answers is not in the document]
danepitkin: Thank you! I had assumed the github action would have used the API under the hood, but I hadn't actually managed to test it yet.

.github/workflows/test.yml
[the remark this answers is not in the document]
danepitkin: I'll remove all `workflow_run` usage. They need a big red security warning in the docs to it's more obvious!

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && startsWith(github.ref, 'refs/tags/') }}
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+          path: dist/
```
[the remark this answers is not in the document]
danepitkin: I'll do that. I was trying to be fancy to keep things DRY

.github/workflows/build.yml
danepitkin: 3.x was also recommended by the official [docs](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/), which is why I chose it.

danepitkin:
I've updated the code to remove `workflow_run` and was able to test the github actions locally for test.yml and the build job in release.yml up until upload_artifact step so far.

assignUser (approved):
LGTM now. A minor change that would mostly be for comparability [future changes](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/) to the default token permissions would be to explicitly set the minimum required permissions: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

gforsyth (approved):
Thanks for putting this together @danepitkin !

And thanks for the security check @assignUser !

@jorisvandenbossche -- do you want to take another look over this?  If not, I'll merge this in later today.

danepitkin:
Thanks all for the review! @gforsyth any recommendations on how to get PyPI secrets configured? Are there substrait-owned secrets we can reuse?

westonpace:
> @westonpace -- I couldn't find a PyPI username for you, if you have one, let me know and I'm happy to add you.

@gforsyth 

I don't have one.  However, if me having one would help in any way then please don't hesitate to reach out and I'll add one.

westonpace:
closes #3

danepitkin:
Do we think its OK merging now, or should we wait for the secrets to be added to the repo?

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
+  publish_test:
+    name: Publish to Test PyPI
+    runs-on: ubuntu-latest
+    needs: build
+    if: github.ref == 'refs/heads/main'
```
jorisvandenbossche, line 72: Do we want to push to the test pypi server for every commit on main?

jorisvandenbossche (approved):
Looks good!

.github/workflows/release.yml
gforsyth: probably not -- but we can leave it in place for the initial commit (after credentials are setup) to make sure everything is working, then strip it out

assignUser: Could be triggered manually via `workflow_dispatch` if that would be useful?

danepitkin: This workflow was recommended by the Python docs[1] themselves: 

> Now, whenever you push a tagged commit to your Git repository remote on GitHub, this workflow will publish it to PyPI. And it’ll publish any push to TestPyPI which is useful for providing test builds to your alpha users as well as making sure that your release pipeline remains healthy!

[1] https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

gforsyth:
I've created the tokens and added them to the repository secrets -- going to merge this and see what happens.

gforsyth:
Looks like the release failed because of some unexpected yaml: https://github.com/substrait-io/substrait-python/actions/runs/5123352674/workflow

gforsyth:
#8 Fixed the yaml issue and the release pipeline ran.  We're getting close!

Now it's that the versioneer style version isn't compliant:
https://github.com/substrait-io/substrait-python/actions/runs/5123895480/jobs/9215062884
```
 ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
         '0.1.dev1+g2e6f6a5' is an invalid value for Version. Error: Can't use  
         PEP 440 local versions. See                                            
         https://packaging.python.org/specifications/core-metadata for more     
         information.  
```

danepitkin:
Added https://github.com/substrait-io/substrait-python/pull/9

PR #9, opened 2023-05-30
fix: package version excludes local version scheme

danepitkin:


danepitkin:
The version now defaults to `0.1.dev4` instead of `0.1.dev4+g2e6f6a5`.

gforsyth (approved):
Thanks @danepitkin !  Let's try this out.

gforsyth:
that did it!
https://test.pypi.org/manage/project/substrait/release/0.1.dev1/

danepitkin:
Woohoo! I'm going to tag `main` with v0.1.0 and see what happens

danepitkin:
I forgot I don't have committer access. @gforsyth could you try `git tag v0.1.0 && git push upstream v0.1.0`?

jorisvandenbossche:
FWIW, I think it is actually fine to keep how it was before: it's a "local version", which is fine for local development, in the idea that the only packages that get normally uploaded to PyPI are always tagged, and so will not have a local identifier in the version (that's how we use this in pandas, for example)

(anyway, doesn't really matter either way, and I suppose if we want to keep using testpypi, this is needed)

danepitkin:
> FWIW, I think it is actually fine to keep how it was before: it's a "local version", which is fine for local development, in the idea that the only packages that get normally uploaded to PyPI are always tagged, and so will not have a local identifier in the version (that's how we use this in pandas, for example)
> 
> (anyway, doesn't really matter either way, and I suppose if we want to keep using testpypi, this is needed)

Hmm, I wonder if there is a way to enable only for PyPI uploads..

gforsyth:
I'd be more inclined to drop the `testpypi` uploads and re-enable the local version (after a few iterations).  If we find that we are often releasing pre-releases, we can revisit that.

gforsyth:
https://pypi.org/project/substrait/0.1.0/

jorisvandenbossche:
>  If we find that we are often releasing pre-releases, we can revisit that.

And pre-releases (if we tag with alpha or beta) can also go on the main PyPI

PR #14, opened 2023-06-15
feat: add upstream substrait version to package

### LINKED ISSUE #11: feat: add upstream substrait version to package
There's the version of `substrait` the Python package and the version of `substrait` the spec / proto implementation.

Maybe we can add `__substrait_version__` and have that refer to the version of the submodule?

gforsyth:
resolves #11

### DIFF
--- a/src/substrait/__init__.py
+++ b/src/substrait/__init__.py
@@ -2,3 +2,7 @@ try:
     from ._version import __version__
 except ImportError:
     pass
+
+__substrait_version__ = "0.30.0"
+__substrait_hash__ = "3259a1ba"
+__minimum_substrait_version__ = "0.30.0"


richtia (approved):

src/substrait/__init__.py
```diff
@@ -2,3 +2,7 @@
     from ._version import __version__
 except ImportError:
     pass
+
+__substrait_version__ = "0.30.0"
+__substrait_hash__ = "3259a1ba"
+__minimum_substrait_version__ = "0.30.0"
```
jorisvandenbossche, line 8: What does the "minimum" version mean exactly? (is that a concept of substrait itself?)

gforsyth: Looking forward, it's supposed to represent the oldest version of substrait that this library can parse a plan from.  I don't think it will be used right away.

PR #24, opened 2023-09-27
chore: add bash script to handle update

gforsyth:
my bash is way out of date, but this is a simple script to handle
checking out the desired tag in the submodule, updating the version and
hash in the `__init__.py` file, and then running `gen_proto.sh`.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+if [[ -z "$1" ]]; then
+    echo "Usage: bash update.sh <version>" 1>&2
+    echo ""
+    echo "Example: bash update.sh v0.34.0" 1>&2
+    exit 1
+fi
+
+VERSION=$1
+
+DIR=$(cd $(dirname $0) && pwd)
+pushd "${DIR}"/third_party/substrait/
+git checkout $VERSION
+SUBSTRAIT_HASH=$(git rev-parse --short HEAD)
+popd
+
+VERSION=$(echo $VERSION | sed 's/v//')
+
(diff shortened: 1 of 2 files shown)


update.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+if [[ -z "$1" ]]; then
+    echo "Usage: bash update.sh version" 1>&2
```
EpsilonPrime, line 4: I put replaceable arguments in angle brackets (`<version>`) but that's a personal style.  Up to you either way.

EpsilonPrime (approved):
The main thing to watch about bash is making sure that anything that calls out is properly escaped or quoted and that is being done here.

update.sh
```diff
@@ -0,0 +1,22 @@
...
+    echo ""
+    echo "Example: bash update.sh v0.34.0" 1>&2
+    exit 1
+fi
+
+VERSION=$1
+
+pushd third_party/substrait/
```
mbrobbel, line 12: This assumes that the working directory is the same as the script file location. You can make calling this script from any location work by adding:
```suggestion
DIR=$(cd $(dirname $0) && pwd)
pushd "${DIR}"/third_party/substrait/
```

update.sh
gforsyth: That's a good call. Added.

update.sh
gforsyth: Nice improvement, thanks @mbrobbel !

EpsilonPrime (approved):
lgtm

PR #28, opened 2023-10-10
chore(bash): update script per shellcheck recs

gforsyth:
Periodically I remember that `shellcheck` exists.  Today was one of those days and so I've made a 
few small syntax changes to the update script to avoid some common bashy gotchas.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -12,13 +12,13 @@ VERSION=$1
 echo "Updating substrait submodule..."
 git submodule update --remote third_party/substrait
 
-DIR=$(cd $(dirname $0) && pwd)
-pushd "${DIR}"/third_party/substrait/
-git checkout $VERSION
+DIR=$(cd "$(dirname "$0")" && pwd)
+pushd "${DIR}"/third_party/substrait/ || exit
+git checkout "$VERSION"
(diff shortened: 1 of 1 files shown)


EpsilonPrime (approved):

update.sh
```diff
@@ -12,13 +12,13 @@ VERSION=$1
...
-git checkout $VERSION
+DIR=$(cd "$(dirname "$0")" && pwd)
+pushd "${DIR}"/third_party/substrait/ || exit
+git checkout "$VERSION"
 SUBSTRAIT_HASH=$(git rev-parse --short HEAD)
-popd
+popd || exit
 
-VERSION=$(echo $VERSION | sed 's/v//')
+VERSION=${VERSION//v/}
```
EpsilonPrime, line 21: Does this add a v to the beginning or remove it?

EpsilonPrime (approved):

update.sh
gforsyth: it removes it -- bash substitution uses a different collection of slashes as compared to `sed` because of course it does.

`${variable//find/replace}`

richtia (approved):

PR #35, opened 2023-11-28
chore(substrait): bump to v0.39.0

gforsyth:
Note, this contains a breaking change in map behavior.

See https://github.com/substrait-io/substrait/releases/tag/v0.39.0 for
more info.

### DIFF
--- a/src/substrait/__init__.py
+++ b/src/substrait/__init__.py
@@ -3,6 +3,6 @@ try:
 except ImportError:
     pass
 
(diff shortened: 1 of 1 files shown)


EpsilonPrime (approved):
Technically the map change is a merely in semantics, not implementation.  So code won't break.  And as long as you're using it with two known systems nothing will break.

gforsyth:
Good to know, thanks!

PR #45, opened 2024-02-06
feat:  add a module to read all Substrait plan formats

EpsilonPrime:
With the generated Substrait plan protobufs already in this repository it is
possible to read and write binary protobufs, text protobufs, and json-encoded
protobufs.  This PR adds the capability to additionally read and write the
Substrait text format as well as auto-detect formats when reading.  It does this
by wrapping the substrait-cpp implementation with ctypes.


EpsilonPrime:
This PR depends on https://github.com/substrait-io/substrait-cpp/pull/91 (the substrait-cpp submodule dependency). As concern will be to keep the version of Substrait that both substrait-python and substrait-cpp depend on in sync.

tests/test_planloader.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+from substrait.planloader import planloader
+
+
+def test_main():
+    print(planloader.__file__)
+    print(dir(planloader))
```
westonpace, line 9: Remove?

tests/test_planloader.py
```diff
@@ -0,0 +1,11 @@
...
+    testplan = planloader.load_substrait_plan('tests/tpch-plan01.json')
+    planloader.save_substrait_plan(testplan, 'myoutfile.splan', planloader.PlanFileFormat.TEXT.value)
```
westonpace, line 11: Round trip?  Or would the resulting be different?

Probably not very important since this is just testing that we can call the method and it doesn't explode.

src/substrait/planloader/planloader.py
```diff
@@ -0,0 +1,82 @@
...
+        A Plan protobuf object if successful.
+    Raises:
+        PlanFileException if an except occurs while converting or reading from disk.
+    """
+    result = external_load_substrait_plan(filename.encode('UTF-8'))
+    if result.contents.errorMessage:
+        raise PlanFileException(result.contents.errorMessage)
+    data = ctypes.string_at(result.contents.buffer, result.contents.size)
+    plan = plan_pb2.Plan()
+    plan.ParseFromString(data)
```
westonpace, line 67: Is there any chance something goes wrong here that doesn't free the plan?  For example, maybe the plan fails to deserialize into protobuf?  Can you use a try / finally?

westonpace (approved):
Let's avoid the potential memory leak and then this is good from my perspective.  Can't comment too much on the build.

gforsyth:
Thanks for putting this together, @EpsilonPrime !  Is the text format documented somewhere?  

I have a few questions:
1. Can we use something like `pybind11` or `nanobind` instead of bare `ctypes`?
2. How complicated is the text format?  It might make more sense to reimplement the parser in Python for the Python library, just for packaging and testing reasons, because...

we currently ship a single no-arch wheel. Once we're wrapping c++ we'll have to build versions for every architecture and also for every version of Python we aim to support.

EpsilonPrime:
> Thanks for putting this together, @EpsilonPrime ! Is the text format documented somewhere?
> 
> I have a few questions:
> 
> 1. Can we use something like `pybind11` or `nanobind` instead of bare `ctypes`?
> 2. How complicated is the text format?  It might make more sense to reimplement the parser in Python for the Python library, just for packaging and testing reasons, because...
> 
> we currently ship a single no-arch wheel. Once we're wrapping c++ we'll have to build versions for every architecture and also for every version of Python we aim to support.

I'll look into the other ways of including the C++ code -- performance really isn't a concern though because we have to serialize and deserialize the plan anyway to cross the language boundary (unless we want to enforce C++ protobuffers in Python which seems even more restrictive since we'd need to match versions more closely).

While implementing in native Python would be nice there's a full parser behind the text format which required 6 months of implementation in C++ (take a look at [the substrait-io/substrait-cpp change log](https://github.com/substrait-io/substrait-cpp/commits/main/?after=66c8d9716036301cf89e3a870837c93536b31d9d+34)).  There is an Antlr4 grammar file which specifies the language for the text format so part of the work is available and perhaps it'd be easier to power through it given that groundwork.  My availability is limited (probably no time for a second language version this year) so if it's possible in the short term, that'd be ideal.

One thing we may want to consider is temporarily making the new library hidden behind a flag so that systems not ready to take on the feature don't need to be burdened by an OS-specific build.

tests/test_planloader.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+from substrait.planloader import planloader
+
+
+def test_main():
+    print(planloader.__file__)
+    print(dir(planloader))
+    testplan = planloader.load_substrait_plan('tests/tpch-plan01.json')
```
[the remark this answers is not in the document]
gforsyth: Testing stuff is great, but I would be concerned about having a test fail in `substrait-python` that then requires upstream changes in either `substrait-cpp` or `consumer-testing`

tests/test_planloader.py
EpsilonPrime: Done.

tests/test_planloader.py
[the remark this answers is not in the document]
EpsilonPrime: The thought here is that we're testing the connecting the connectivity to the textplan library in C++.  The substrait-cpp has extensive tests including round trip tests of all of the TPCH queries.

tests/test_planloader.py
EpsilonPrime: The round trip would be vulnerable to upstream changes in format.  As you note the connectivity test here is likely sufficient.

danepitkin:
Very cool! It's great to see new features added to the python impl.

I'm a big +1 on using `nanobind` (1st choice) or `pybind11`. 

Also, it might be nice to make these additional features optional to downstream dependencies. For example, would Ibis want this included by default? We could organize the python package such that `substrait` only includes the generated protobuf classes and `substrait[serializers]`(?) includes the binary/text/json [de]serializers. And then `substrait[all]` will always include everything. WDYT?

jorisvandenbossche:
> We could organize the python package such that `substrait` only includes the generated protobuf classes and `substrait[serializers]`(?) includes the binary/text/json [de]serializers. And then `substrait[all]` will always include everything. WDYT?

AFAIK that will only work if we package the python bindings for substrait-cpp separately (a separate python package), such that substrait-python can depend on that additional package (I think extras can only be used for dependencies, not for enabling features within the library itself, except by relying on an external dependency)

.github/workflows/release.yml
```diff
@@ -25,6 +25,14 @@ jobs:
...
         run: |
           python -m pip install build --user
+      - name: Build Substrait planloader library
+        run: |
+          cd ${{ github.workspace }}/third_party/substrait-cpp
+          make release
+      - name: Install Substrait planloader library
+        run: |
+          cd ${{ github.workspace }}/third_party/substrait-cpp/build-Release/export/planloader
+          make install
```
gforsyth, line 35: We'll want to handle compilation via the `pyproject.toml` file, I'll leave some notes there.

.github/workflows/test.yml
```diff
@@ -26,6 +26,14 @@ jobs:
...
         with:
           python-version: ${{ matrix.python }}
+      - name: Build Substrait planloader library
+        run: |
+          cd ${{ github.workspace }}/third_party/substrait-cpp
+          make release
+      - name: Install Substrait planloader library
+        run: |
+          cd ${{ github.workspace }}/third_party/substrait-cpp/build-Release/export/planloader
+          make install
```
gforsyth, line 36: We'll want to handle compilation via the `pyproject.toml` file, I'll leave some notes there.

gforsyth (commented):
In the `pyproject.toml` file we _might_ be able to get this working if we add `cmake` to the `requires` list under `[build-system]` -- I believe that the `setuptools.build_meta` `build-backend` has at least some support for that.  
If not, we can look at `scikit-build` to help with getting things building as part of the `pip` installation, and that will help us feed something like `cibuildwheel` for handling packaging for all the python and architecture versions.

gforsyth:
In the event that the nested submodules make this extra painful (which seems... likely?) we could always use `git subtree`, which has much more predictable behavior in my estimation.

amol-:
@EpsilonPrime can you please add an example to the README to show how to parse/produce a text plan?

EpsilonPrime:
I'm looking at making this a separate library.

PR #46, opened 2024-02-08
feat: add a module to read all Substrait plan formats (pybind)

EpsilonPrime:


EpsilonPrime:
This isn't simpler than the ctypes approach so I'm abandoning it.

PR #50, opened 2024-03-13
feat: add substrait.proto convenience module and document it

amol-:
Add a ``substrait.proto`` module that gives access to the Substrait protocol classes 
removing the need to navigate the hierarchy automatically generated by protobuf. 
It also provides access to the modules without the ``_pb2`` suffix 
which is an implementation detail of the protobuf version used.

Provides examples on how to generate and read back Substrait plans 
using the substrait-python module itself.


tests/test_proto.py
```diff
@@ -12,3 +12,21 @@ def test_imports():
     from substrait.gen.proto.type_expressions_pb2 import DerivationExpression
     from substrait.gen.proto.type_pb2 import Type
     from substrait.gen.proto.extensions.extensions_pb2 import SimpleExtensionURI
+
```
danepitkin, line 15: We can go ahead and delete the test_imports() function since the new test indirectly tests these imports.

danepitkin (approved):
LGTM! Thanks @amol- this is already much cleaner for the user :)

tests/test_proto.py
```diff
@@ -12,3 +12,21 @@ def test_imports():
...
+
+def test_proto_proxy_module():
+    """Test that protocol classes are made available in substrait.proto"""
+    import substrait.proto
+
+    assert {"Plan", "Type", "NamedStruct", "RelRoot"} & set(dir(substrait.proto))
```
gforsyth, line 21:
```suggestion
    assert {"Plan", "Type", "NamedStruct", "RelRoot"} <= set(dir(substrait.proto))
```

I'm assuming here that the goal is checking that these know values are a subset of the contents of `substrait.proto`?

tests/test_proto.py
```diff
@@ -12,3 +12,21 @@ def test_imports():
...
+        "algebra",
+        "capabilities",
+        "extensions",
+        "extended_expression",
+        "function",
+        "parameterized_types",
+        "plan",
+        "type_expressions",
+        "type",
+    } & set(dir(substrait.proto))
```
gforsyth, line 32:
```suggestion
    } <= set(dir(substrait.proto))
```

gforsyth (commented):
looks good overall -- two small suggestions for the tests.

EpsilonPrime:
So the purpose of this PR is to get the protos exposed with the simpler package name without changing the fact that they are registered with the existing one?

One reason the current package name is used is to prevent conflicts when the C++ version of the protos is also loaded under ::substrait::proto.  We should probably document this desire along with the fact that we aren't simply changing what package we generate the protos in.

tests/test_proto.py
amol-: Ops, good catch, it would have passed even if only one existed

tests/test_proto.py
amol-: I think there is a value in keeping this test to ensure backward compatibility.
`substrait.proto` might continue to work but import the classes from modules with different names than the current ones. 

In such case we would be breaking backward compatibility for anyone directly accessing the classes from the `_pb2` modules and that's something we want to catch and be aware of.

amol-:
> So the purpose of this PR is to get the protos exposed with the simpler package name without changing the fact that they are registered with the existing one?
> 

Correct, I wanted to retain backward compatibility for anyone importing the classes from the autogenerated modules, but also make the automatic generation an implementation detail.

> One reason the current package name is used is to prevent conflicts when the C++ version of the protos is also loaded under ::substrait::proto. We should probably document this desire along with the fact that we aren't simply changing what package we generate the protos in.

Can you elaborate? How does the C++ library namespace influence the Python namespace?

EpsilonPrime:
Depending on your options you can actually use C++ protos for the Python under the hood.  But ignoring that you can still end up with two definitions for protobufs which will collide and cause an exception when the module loads.  If they are in different namespaces that collision won't happen (you can pretend to have two different sets of protos).

On a side note, I have been having issues with the current protos (lint and IDE) because the attributes aren't available for the package until runtime.  The solution I used for the spark connect protos was to use mypy-protobuf (add --mypy_out to the protoc call in gen_proto.sh) which created pyi files which also would need to be checked in.  Would this approach work here?

EpsilonPrime (approved):

gforsyth (approved):
thanks @amol!

PR #52, opened 2024-03-27
feat: add interface files with mypy information to aid with type checking

EpsilonPrime:


gforsyth (approved):
this seems like a reasonable change.  I think you also need to add an empty file called `py.typed` in the `src/` directory to inform downstream projects that type information is available.

EpsilonPrime:
> this seems like a reasonable change. I think you also need to add an empty file called `py.typed` in the `src/` directory to inform downstream projects that type information is available.

Having the type information and enforcing it seem like two different levels.  I do want to add `py.typed` but I believe that should be done in a followup PR especially as it has the possibility to affect downstream packages.

PR #53, opened 2024-04-10
feat: python native JSON format parser

amol-:
Add `substrait.json.load_json` and `substrait.json.parse_json` functions able to load
 the JSON representation of a Substrait plan to a `substrait.proto.Plan` object.

It also adds the `substrait-cpp` repository as a git submodule to reuse the test files.
This is reasonable because we might end up using the cpp library in the future 
to create bindings to other features too, so it's helpful to already have it as a submodule.


EpsilonPrime:
You can also use the Protobuf json_format module to load/save these files directly:

https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html

src/substrait/json.py
```diff
@@ -0,0 +1,76 @@
...
+        return {literal_type: dict(**literal_def, value=base64.b64decode(literal_value))}
+    return literal
+
+
+def _fix_fetch(jsonobj):
+    """Fix offset and count in fetch definitions.
+
+    For some reason substrait producers are generating
+    fetch with offset and count being strings,
+    while their defintion in the proto is as int64.
```
mbrobbel, line 65:
```suggestion
    while their definition in the proto is as int64.
```

amol-:
> You can also use the Protobuf json_format module to load/save these files directly:
> 
> https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html

That's great, I see it also deals with the keyname transposition `preserving_proto_field_name`, do you know if it handles the base64 translation too?

amol-:
> Ok, I'll try to switch the implementation to use `google.protobuf.json_format.Parse` and see if anything breaks

Moved to using `json_format.Parse`, all tests passing like before

.github/workflows/test.yml
```diff
@@ -32,4 +32,4 @@ jobs:
           python -m pip install ".[test]"
       - name: Run tests
         run: |
-          python -m pytest
+          python -m pytest tests
```
EpsilonPrime, line 35: If you don't want to change the way pytest is called you can add the following to pyproject.toml in the tool.pytest.ini.options section:

`addopts = "--ignore=third_party"
`

src/substrait/json.py
```diff
@@ -0,0 +1,25 @@
+from google.protobuf import json_format
```
EpsilonPrime, line 1: There are three Google ways of loading/saving protobuffers:

binary
JSON format
text format

It's worth considering how you want to handle the four different types (including the Substrait text format).  The C++ package implements methods to read/write any.

tests/test_json.py
```diff
@@ -0,0 +1,72 @@
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        filename = pathlib.Path(tmpdir) / "jsonfile.json"
+        write_json(parsed_plan, filename)
+        assert load_json(filename) == parsed_plan
+
+
+def _strip_json_comments(jsonfile):
+    # The JSON files in the cpp testsuite are prefixed with
+    # a comment containing the SQL that matches the json plan.
+    # As Python JSON parser doesn't support comments,
```
EpsilonPrime, line 70: Technically none of the parsers support the comments including the json_format library.

update_proto.sh
EpsilonPrime, line 1: There's documentation pointing to this file that should be updated.

tests/test_json.py
```diff
@@ -0,0 +1,72 @@
...
+    pathlib.Path(os.path.dirname(__file__))
+    / ".."
+    / "third_party"
+    / "substrait-cpp"
+    / "src"
+    / "substrait"
+    / "textplan"
+    / "data"
+)
+JSON_TEST_FILE = list(JSON_FIXTURES.glob("*.json"))
```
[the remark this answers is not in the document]
amol-: I don't think the order of execution matters, but happy to make it sorted

tests/test_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+from substrait.proto import Plan
+from substrait.json import load_json, parse_json, dump_json, write_json
+
+import pytest
+
+
+JSON_FIXTURES = (
+    pathlib.Path(os.path.dirname(__file__))
+    / ".."
```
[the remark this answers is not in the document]
amol-: I preferred to go for `/ ".."` because I find it more readable than invoking methods when expressing a path, but happy to go for `parent` if there is a preference

tests/test_json.py
```diff
@@ -0,0 +1,72 @@
...
+    / "third_party"
+    / "substrait-cpp"
+    / "src"
+    / "substrait"
+    / "textplan"
+    / "data"
+)
+JSON_TEST_FILE = list(JSON_FIXTURES.glob("*.json"))
+JSON_TEST_FILENAMES = [path.name for path in JSON_TEST_FILE]
```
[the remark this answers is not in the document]
amol-: I think that having the explicit filename here makes a bit more obvious what the test is using

.github/workflows/test.yml
amol-: You are right, this should probably go into `pyproject.toml` as it's a good idea locally too. 
I'll specify `testpaths` there

update_proto.sh
amol-: Ah thanks, do you have a link to it? I grepped for `update.sh` when I did the name change but couldn't find any reference.

EpsilonPrime: Found it -- it is called "upgrade.sh" in CONTRIBUTING.md.

tests/test_json.py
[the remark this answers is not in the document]
EpsilonPrime: Fine with me.

tests/test_json.py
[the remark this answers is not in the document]
EpsilonPrime: Fine with me.

update_proto.sh
amol-: Ah, that explains why I couldn't find it. The scriptname in CONTRIBUTING.md was wrong, it's `upgrade.sh` while the correct name actually `update.sh`. Fixed it and updated to `update_proto.sh`

EpsilonPrime (approved):

amol-:
Is this ok to merge? I'd like to use this for further work I'm planning to contribute to Substrait so it would be helpful to have it published

amol-:
> Not a blocker, but adding the `substrait-cpp` repo as a submodule just for the json test files feels a little bloated.

I could have copied the data, but given that we are already planning to add it as a dependency for textplan, it made sense to avoid a copy and use the test files directly from the submodule

gforsyth:
> I could have copied the data, but given that we are already planning to add it as a dependency for textplan, it made sense to avoid a copy and use the test files directly from the submodule

That's reasonable.  I prefer `subtree` to `submodule` but I think that ship has sailed.

gforsyth (approved):

PR #68, opened 2025-03-02
feat: add registry for extension functions

tokoko:
adds FunctionRegistry that handles lookup and type inference of extension functions.
uses derivation expressions under the hood to generate return types.
pyyaml is now part of `extensions` extra.

### DIFF
--- a/src/substrait/function_registry.py
+++ b/src/substrait/function_registry.py
@@ -50,9 +50,6 @@ def normalize_substrait_type_names(typ: str) -> str:
     return typ
 
 
-id_generator = itertools.count(1)
-
-
 def to_integer_option(txt: str):
     if txt.isnumeric():
         return ParameterizedType.IntegerOption(literal=int(txt))
@@ -153,7 +150,7 @@ def to_parameterized_type(dtype: str):
             user_defined=ParameterizedType.ParameterizedUserDefined()
         )
     else:
-        raise Exception(f"Unkownn type - {dtype}")
+        raise Exception(f"Unknown type - {dtype}")
 
 
 def violates_integer_option(
@@ -210,11 +207,11 @@ def covers(dtype: Type, parameterized_type: ParameterizedType, parameters: dict)
 
 
 class FunctionEntry:
-    def __init__(self, uri: str, name: str, impl: Mapping[str, Any]) -> None:
+    def __init__(self, uri: str, name: str, impl: Mapping[str, Any], anchor: int) -> None:
         self.name = name
         self.normalized_inputs: list = []
         self.uri: str = uri
-        self.anchor = next(id_generator)
+        self.anchor = anchor
         self.arguments = []
         self.rtn = impl["return"]
         self.nullability = impl.get("nullability", False)
@@ -286,7 +283,7 @@ class FunctionRegistry:
         for named_functions in definitions.values():
             for function in named_functions:
                 for impl in function.get("impls", []):
-                    func = FunctionEntry(uri, function["name"], impl)
+                    func = FunctionEntry(uri, function["name"], impl, next(self.id_generator))
                     if (
                         func.uri in self._function_mapping
                         and function["name"] in self._function_mapping[func.uri]
--- a/src/substrait/function_registry.py
+++ b/src/substrait/function_registry.py
@@ -317,18 +317,19 @@ class FunctionEntry:
 
 
 class FunctionRegistry:
-    def __init__(self) -> None:
+    def __init__(self, load_default_extensions=True) -> None:
         self._function_mapping: dict = defaultdict(dict)
-        self.id_generator = itertools.count(1)
+        self._id_generator = itertools.count(1)
 
-        self.uri_aliases = {}
+        self._uri_aliases = {}
 
-        for fpath in importlib_files("substrait.extensions").glob(  # type: ignore
-            "functions*.yaml"
-        ):
-            uri = f"{DEFAULT_URI_PREFIX}/{fpath.name}"
-            self.uri_aliases[fpath.name] = uri
-            self.register_extension_yaml(fpath, uri)
+        if load_default_extensions:
+            for fpath in importlib_files("substrait.extensions").glob(  # type: ignore
+                "functions*.yaml"
+            ):
+                uri = f"{DEFAULT_URI_PREFIX}/{fpath.name}"
+                self._uri_aliases[fpath.name] = uri
+                self.register_extension_yaml(fpath, uri)
 
     def register_extension_yaml(
         self,
@@ -346,7 +347,7 @@ class FunctionRegistry:
             for function in named_functions:
                 for impl in function.get("impls", []):
                     func = FunctionEntry(
-                        uri, function["name"], impl, next(self.id_generator)
+                        uri, function["name"], impl, next(self._id_generator)
                     )
                     if (
                         func.uri in self._function_mapping
@@ -360,7 +361,7 @@ class FunctionRegistry:
     def lookup_function(
         self, uri: str, function_name: str, signature: tuple
     ) -> Optional[tuple[FunctionEntry, Type]]:
-        uri = self.uri_aliases.get(uri, uri)
+        uri = self._uri_aliases.get(uri, uri)
 
         if (
             uri not in self._function_mapping
--- a/src/substrait/function_registry.py
+++ b/src/substrait/function_registry.py
@@ -1,0 +1,316 @@
+from substrait.gen.proto.parameterized_types_pb2 import ParameterizedType
+from substrait.gen.proto.type_pb2 import Type
+from importlib.resources import files as importlib_files
+import itertools
+from collections import defaultdict
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Any, Optional, Union
+from .derivation_expression import evaluate
+
+import yaml
+import re
+
+_normalized_key_names = {
+    "binary": "vbin",
+    "interval_compound": "icompound",
+    "interval_day": "iday",
+    "interval_year": "iyear",
+    "string": "str",
+    "timestamp": "ts",
+    "timestamp_tz": "tstz",
+}
+
+
+def normalize_substrait_type_names(typ: str) -> str:
+    # First strip off any punctuation
+    typ = typ.strip("?").lower()
+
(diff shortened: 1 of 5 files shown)


tokoko:
This is currently blocked by #69

src/substrait/function_registry.py
```diff
@@ -0,0 +1,316 @@
...
+    "interval_day": "iday",
+    "interval_year": "iyear",
+    "string": "str",
+    "timestamp": "ts",
+    "timestamp_tz": "tstz",
+}
+
+
+def normalize_substrait_type_names(typ: str) -> str:
+    # First strip off any punctuation
```
EpsilonPrime, line 26: Question marks are the nullability for a given type.  I'd call that out explicitly.

src/substrait/function_registry.py
```diff
@@ -0,0 +1,316 @@
...
+        if typ.lower().startswith(complex_type):
+            typ = abbr
+
+    # Then pass through the dictionary of mappings, defaulting to just the
+    # existing string
+    typ = _normalized_key_names.get(typ.lower(), typ.lower())
+    return typ
+
+
+id_generator = itertools.count(1)
```
EpsilonPrime, line 53: For long running programs this id generation would get pretty big.  Would it be better to reset for each plan (perhaps requiring a function registry class)?

src/substrait/function_registry.py
```diff
@@ -0,0 +1,316 @@
...
+    elif dtype.startswith("any"):
+        return ParameterizedType(
+            type_parameter=ParameterizedType.TypeParameter(name=dtype)
+        )
+    elif dtype.startswith("u!") or dtype == "geometry":
+        return ParameterizedType(
+            user_defined=ParameterizedType.ParameterizedUserDefined()
+        )
+    else:
+        raise Exception(f"Unkownn type - {dtype}")
```
EpsilonPrime, line 156: Unknown

src/substrait/function_registry.py
```diff
@@ -0,0 +1,316 @@
...
+    def __init__(self) -> None:
+        self._function_mapping: dict = defaultdict(dict)
+        self.id_generator = itertools.count(1)
+
+        self.uri_aliases = {}
+
+        for fpath in importlib_files("substrait.extensions").glob(  # type: ignore
+            "functions*.yaml"
+        ):
+            uri = f"https://github.com/substrait-io/substrait/blob/main/extensions/{fpath.name}"
```
EpsilonPrime, line 270: If someone has their own set of extension files they would be unable to use this routine.

src/substrait/function_registry.py
```diff
@@ -0,0 +1,316 @@
...
+class FunctionRegistry:
+    def __init__(self) -> None:
+        self._function_mapping: dict = defaultdict(dict)
+        self.id_generator = itertools.count(1)
+
+        self.uri_aliases = {}
+
+        for fpath in importlib_files("substrait.extensions").glob(  # type: ignore
+            "functions*.yaml"
+        ):
```
EpsilonPrime, line 269: Consider making a constant for the default URI prefix.

src/substrait/function_registry.py
tokoko: oops, my bad. this should not be here. `id_generator` is also being created in `FunctionRegistry` constructor, that's the one that's supposed to be used instead. That way it will be up to the application itself whether it wants to reuse a single FunctionRegistry (shared id generation) or create a brand new one for each plan.

src/substrait/function_registry.py
tokoko: This is only meant to preload extensions from substrait repo by default for convenience, they would still be able to ignore these and call `register_extension_yaml` method for their own extension files instead. Do you mean that they would waste resources by loading default extensions if they know they won't use it? I guess we can make it configurable in the constructor:

```
def __init__(self, load_default_extensions: bool = True)
```

src/substrait/function_registry.py
tokoko: done

src/substrait/function_registry.py
```diff
@@ -0,0 +1,340 @@
...
+    "precision_timestamp_tz": "ptstz",
+    "struct": "struct",
+    "list": "list",
+    "map": "map",
+}
+
+
+def normalize_substrait_type_names(typ: str) -> str:
+    # First strip nullability marker
+    typ = typ.strip("?").lower()
```
EpsilonPrime, line 52: The nullability marker doesn't necessarily go at the end.  Will this work as expected for `struct?<str,str>`?

EpsilonPrime: Here are a set of type test cases that might be useful:   https://github.com/substrait-io/substrait-cpp/blob/main/src/substrait/type/tests/TypeTest.cpp

tokoko: thanks, I was convinced it was the other way around. I'll have to handle that in `to_parameterized_type` method as well, then.

tokoko: Made most of the changes, also added proper handling of scalar function's `nullability` field. Some stuff still needs a bit of a rework, but I added a bunch of TODOs for those.

src/substrait/function_registry.py
```diff
@@ -0,0 +1,378 @@
...
+    typ = typ.split("<")[0]
+    # First strip nullability marker
+    typ = typ.strip("?").lower()
+
+    if typ.startswith("any"):
+        return "any"
+    elif typ.startswith("u!"):
+        return typ
+    else:
+        return _normalized_key_names[typ]
```
EpsilonPrime, line 61: What happens if the short name is used instead?  Consider throwing a specific exception here so it's more clear what an end user should do.

src/substrait/function_registry.py
```diff
@@ -0,0 +1,378 @@
...
+        return ParameterizedType.IntegerOption(literal=int(txt))
+    else:
+        return ParameterizedType.IntegerOption(
+            parameter=ParameterizedType.IntegerParameter(name=txt)
+        )
+
+
+# TODO try using antlr grammar here as well
+def to_parameterized_type(dtype: str):
+    if "?" in dtype:
```
EpsilonPrime, line 75: A non-nullable struct with a nullable field would end up as a nullable struct with a non-nullable field possibly?

src/substrait/function_registry.py
```diff
@@ -0,0 +1,378 @@
...
+        return ParameterizedType(i32=Type.I32(nullability=nullability))
+    elif dtype == "i64":
+        return ParameterizedType(i64=Type.I64(nullability=nullability))
+    elif dtype == "fp32":
+        return ParameterizedType(fp32=Type.FP32(nullability=nullability))
+    elif dtype == "fp64":
+        return ParameterizedType(fp64=Type.FP64(nullability=nullability))
+    elif dtype == "timestamp":
+        return ParameterizedType(timestamp=Type.Timestamp(nullability=nullability))
+    elif dtype == "timestamp_tz":
```
EpsilonPrime, line 97: Consider creating an enum/type/constant for the types in the future so that there is only one place where the names are.  (Not this PR)

src/substrait/function_registry.py
```diff
@@ -0,0 +1,378 @@
...
+    elif dtype.startswith("precision_timestamp"):
+        (_, precision, _) = re.split(r"\W+", dtype)
+
+        return ParameterizedType(
+            precision_timestamp=ParameterizedType.ParameterizedPrecisionTimestamp(
+                precision=to_integer_option(precision), nullability=nullability
+            )
+        )
+    elif dtype.startswith("precision_timestamp_tz"):
+        (_, precision, _) = re.split(r"\W+", dtype)
```
EpsilonPrime, line 134: If the type is malformed what happens?

src/substrait/function_registry.py
tokoko: yes, I'll give using antlr parser a try rather than reinvent the wheel.

src/substrait/function_registry.py
tokoko: Is using short names in extension yamls legal? I changed dict search to throw a specific exception.

src/substrait/function_registry.py
tokoko: I think I did an okay job of it for now, please take a look.

src/substrait/function_registry.py
tokoko: Got rid of this part of the code while rewriting with antlr, but will take it into consideration.

src/substrait/function_registry.py
tokoko: Removed the function, with the current impl users will probably see an antlr error

src/substrait/function_registry.py
EpsilonPrime: A quick glance of functions_arithmetic.yaml contains a lot of i8 and i16 arguments.  If you put together a test which verifies that you pass all of the core YAML extensions you should be good to go.

My general practice when implementing a spec is to be exact as possible on output and lenient on input.  But since we have the nearly full test set that should be enough.

src/substrait/function_registry.py
EpsilonPrime: If someone had their own directory of YAML files and wanted to use this routine they would instead have to do their own directory visiting sequence.  Not a big deal but if we start going down the dialect path more (as discussed at the community meeting) this may be more common in the future (which means this is fine for now).

EpsilonPrime (approved):
Thanks for all the work!

I'm okay with putting this PR in as is but I'd like to see the test that runs against the extensions in Substrait Core to shake out any issues.

src/substrait/function_registry.py
tokoko: Understood. The reason I went with `register_extension_yaml` and `register_extension_dict` only was because I needed them for default extensions and tests, respectively. Beyond that, it's a bit unclear right now whether people will have a local directory of yamls available or maybe just a collection of http paths, so I decided to leave it up to the users. Having said that, we can always add other helper methods to make popular cases convenient, of course.

src/substrait/function_registry.py
tokoko: instantiating the registry loads all extensions by default, so they are all implicitly being tested. I'll work on more explicit tests, as well, though

PR #71, opened 2025-03-15
feat: extended expression builders

tokoko:
Adds extended expression builder functions: `column`, `literal` and `scalar_function`

### DIFF
--- a/src/substrait/extended_expression.py
+++ b/src/substrait/extended_expression.py
@@ -1,0 +1,32 @@
+import itertools
+import substrait.gen.proto.algebra_pb2 as stalg
+import substrait.gen.proto.type_pb2 as stp
+import substrait.gen.proto.extended_expression_pb2 as stee
+from substrait.utils import type_num_names
+
+
+def column(name: str):
+    def resolve(base_schema: stp.NamedStruct) -> stee.ExtendedExpression:
+        column_index = list(base_schema.names).index(name)
+        lengths = [type_num_names(t) for t in base_schema.struct.types]
+        flat_indices = [0] + list(itertools.accumulate(lengths))[:-1]
+
+        return stee.ExtendedExpression(
+            referred_expr=[
+                stee.ExpressionReference(
+                    expression=stalg.Expression(
+                        selection=stalg.Expression.FieldReference(
+                            root_reference=stalg.Expression.FieldReference.RootReference(),
+                            direct_reference=stalg.Expression.ReferenceSegment(
+                                struct_field=stalg.Expression.ReferenceSegment.StructField(
+                                    field=flat_indices.index(column_index)
+                                )
+                            ),
+                        )
+                    )
+                )
+            ],
+            base_schema=base_schema,
+        )
+
+    return resolve
--- a/src/substrait/extended_expression.py
+++ b/src/substrait/extended_expression.py
@@ -6,11 +6,12 @@ import substrait.gen.proto.extensions.extensions_pb2 as ste
 from substrait.function_registry import FunctionRegistry
 from substrait.utils import type_num_names, merge_extension_uris, merge_extension_declarations
 from substrait.type_inference import infer_extended_expression_schema
-from typing import Callable, Any
+from typing import Callable, Any, Union
 
 UnboundExpression = Callable[[stp.NamedStruct, FunctionRegistry], stee.ExtendedExpression]
 
 def literal(value: Any, type: stp.Type, alias: str = None) -> UnboundExpression:
+    """Builds a resolver for ExtendedExpression containing a literal expression"""
     def resolve(base_schema: stp.NamedStruct, registry: FunctionRegistry) -> stee.ExtendedExpression:
         kind = type.WhichOneof('kind')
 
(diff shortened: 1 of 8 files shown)


src/substrait/extended_expression.py
```diff
@@ -0,0 +1,147 @@
...
+        elif kind == "i64":
+            literal = stalg.Expression.Literal(i64=value, nullable=type.i64.nullability == stp.Type.NULLABILITY_NULLABLE)
+        elif kind == "fp32":
+            literal = stalg.Expression.Literal(fp32=value, nullable=type.fp32.nullability == stp.Type.NULLABILITY_NULLABLE)
+        elif kind == "fp64":
+            literal = stalg.Expression.Literal(fp64=value, nullable=type.fp64.nullability == stp.Type.NULLABILITY_NULLABLE)
+        elif kind == "string":
+            literal = stalg.Expression.Literal(string=value, nullable=type.string.nullability == stp.Type.NULLABILITY_NULLABLE)
+        else:
+            raise Exception(f"Unknown literal type - {type}")
```
EpsilonPrime, line 34: There are also some unimplemented types here (such as dates, times, binary, uuid, etc.).  Is it worth providing guidance here as to how to request/implement support?

src/substrait/extended_expression.py
```diff
@@ -0,0 +1,147 @@
...
+import substrait.gen.proto.extended_expression_pb2 as stee
+import substrait.gen.proto.extensions.extensions_pb2 as ste
+from substrait.function_registry import FunctionRegistry
+from substrait.utils import type_num_names, merge_extension_uris, merge_extension_declarations
+from substrait.type_inference import infer_extended_expression_schema
+from typing import Callable, Any
+
+UnboundExpression = Callable[[stp.NamedStruct, FunctionRegistry], stee.ExtendedExpression]
+
+def literal(value: Any, type: stp.Type, alias: str = None) -> UnboundExpression:
```
EpsilonPrime, line 13: Is there a way to create a literal builder that works for both expressions and extended expressions (avoiding code duplication)?  Perhaps we need to add more support to the other builder parts to accept either.

src/substrait/extended_expression.py
```diff
@@ -0,0 +1,147 @@
...
+                    ),
+                    output_names=[alias if alias else f'literal_{kind}'],
+                )
+            ],
+            base_schema=base_schema,
+        )
+
+    return resolve
+
+def column(name: str):
```
EpsilonPrime, line 50: This is working over a name instead of a column number.  Substrait doesn't need names but I can see this being useful as a convenience.  There should be a version that accepts a field number.

src/substrait/extended_expression.py
EpsilonPrime, line 13: Please add comments for these methods (for instance, note how names are used).

src/substrait/extended_expression.py
```diff
@@ -0,0 +1,147 @@
...
+    def resolve(base_schema: stp.NamedStruct, registry: FunctionRegistry) -> stee.ExtendedExpression:
+        column_index = list(base_schema.names).index(name)
```
EpsilonPrime, line 52: Should this resolution logic be handled in a separate function?

src/substrait/extended_expression.py
```diff
@@ -0,0 +1,147 @@
...
+                    ),
+                    output_names=list(base_schema.names)[names_start:names_end],
+                )
+            ],
+            base_schema=base_schema,
+        )
+
+    return resolve
+
+def scalar_function(uri: str, function: str, *expressions: UnboundExpression, alias: str = None):
```
EpsilonPrime, line 85: Should this accept regular expressions and extended expressions as well?

src/substrait/utils.py
```diff
@@ -0,0 +1,44 @@
...
+        lengths = [type_num_names(t) for t in typ.struct.types]
+        return sum(lengths) + 1
+    elif kind == "list":
+        return type_num_names(typ.list.type)
+    elif kind == "map":
+        return type_num_names(typ.map.key) + type_num_names(typ.map.value)
+    else:
+        return 1
+
+def merge_extension_uris(*extension_uris: Iterable[ste.SimpleExtensionURI]):
```
EpsilonPrime, line 17: A comment explaining what merge does (maintains order but removes duplicates) would be useful here.

src/substrait/extended_expression.py
tokoko: I think there are 3 possible types to be used here: `Expression`, `ExtendedExpression` and `UnboundExpression`. My thinking about pros/cons of each option is:

- The downside of using a regular proto `Expression` type is that `Expression`s can't carry additional information regarding resources that are used in the expression tree, meaning that if you use an extension function or a cte (once this substrait [PR](https://github.com/substrait-io/substrait/pull/794) is merged) that context is thrown away with a regular `Expression` type.

    The same applies to a more Rel-level builders as well, for example a possible join builder can have one of these following signatures: `join(left: Rel, right: Rel, join_type: str, condition: Expression)` or `join(left: Plan, right: Plan, join_type: str, condition: ExtendedExpression)`. I think we should prefer the second one because it allows for each object to independently hold full context of the extenstions/ctes used while building each object and allows them to be merged by the builder function later on.
- `ExtendedExpression` is perfectly okay to use here with the only downside being that in order to build an ExtendedExpression the user needs to be aware of precisely in what context the expression is supposed to be used, in other words once `ExtendedExpression` is built, you've already tied it to a specific Rel schema where it can be used. 
  
- `UnboundExpression` is the currying equivalent of `ExtendedExpression` which allows the user to defer resolving precise Rel schema until later when it's actually used in the context of a specific Rel.

I guess we can probably introduce a type like `Union[Expression, ExtendedExpression, UnboundExpression]` and use it in similar builder signatures, if we want to cover as many use cases as possible. We will have to then do some light type matching on the builder side.

src/substrait/extended_expression.py
tokoko: I'll factor out the logic in a separate util as soon as another use for it appears. Does that sound okay?

src/substrait/extended_expression.py
tokoko: I think a simple Expression builder makes a lot of sense, especially for literals. I'm simply trying to keep things uniform for starters as I don't have a clear idea how to tackle Expression builders for non-literal types.

src/substrait/extended_expression.py
tokoko: that's a good point. I'll follow up with those implementations shortly and let's discuss this in that PR, if that's okay. Honestly I'm not sure how a literal builder for a "complex" type (something like IntervalCompound for example) should look like.

EpsilonPrime (approved):

PR #72, opened 2025-04-12
feat: add builders for aggregate, window funcs and if-then

tokoko:
- moves builders to a new directory
- renames FunctionRegistry to ExtensionRegistry
- adds new extended expression builders


src/substrait/builders/__init__.py
EpsilonPrime, line 1: I think this file was deprecated in Python 3.3.

src/substrait/builders/extended_expression.py
```diff
@@ -0,0 +1,374 @@
...
+                            arguments=[
+                                stalg.FunctionArgument(
+                                    value=e.referred_expr[0].expression
+                                )
+                                for e in bound_expressions
+                            ],
+                            output_type=func[1],
+                        )
+                    ),
+                    output_names=[alias if alias else "scalar_function"],
```
EpsilonPrime, line 161: Should this be a unique name?  Perhaps keep track of generated names and prepend a count for each reuse (and elsewhere).

src/substrait/builders/__init__.py
tokoko: done. I still bump into issues occasionally when init files are missing, but we can always add them if there's a need for it later

src/substrait/builders/extended_expression.py
tokoko: I switched to a rudimentary way of generating names from argument names. (ibis does something like this) Pretty sure it's gonna fail if the output type of an expression contains a struct, but it's a start. I'm not sure if it's worth it to infer an output type just to get the expected cardinality of the `output_names` array.

EpsilonPrime (approved):

PR #73, opened 2025-04-16
feat: add plan builders

tokoko:
- add plan builders for read, project, filter, etc
- add simple builders for types

### DIFF
--- a/src/substrait/builders/plan.py
+++ b/src/substrait/builders/plan.py
@@ -1,6 +1,8 @@
 """
 Plan builders take either Plan or UnboundPlan objects as input rather than plain Rels.
 This is to make sure that additional information like extension types of functions are not lost.
+All builders return UnboundPlan objects that can be materialized to a Plan using an ExtensionRegistry.
+See `examples/builder_example.py` for usage.
 """
 
 from typing import Iterable, Union, Callable
--- a/src/substrait/builders/plan.py
+++ b/src/substrait/builders/plan.py
@@ -1,0 +1,81 @@
+from typing import Iterable, Union
+
+import substrait.gen.proto.algebra_pb2 as stalg
+import substrait.gen.proto.plan_pb2 as stp
+import substrait.gen.proto.type_pb2 as stt
+import substrait.gen.proto.extended_expression_pb2 as stee
+from substrait.extension_registry import ExtensionRegistry
+from substrait.builders.extended_expression import UnboundExtendedExpression
+from substrait.type_inference import infer_plan_schema
+from substrait.utils import merge_extension_declarations, merge_extension_uris
+
+
+def _merge_extensions(*objs):
+    return {
+        "extension_uris": merge_extension_uris(*[b.extension_uris for b in objs]),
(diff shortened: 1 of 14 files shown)


src/substrait/builders/plan.py
```diff
@@ -0,0 +1,294 @@
...
+
+        return stp.Plan(
+            relations=[stp.PlanRel(root=stalg.RelRoot(input=rel, names=ns.names))],
+            **_merge_extensions(bound_left, bound_right, bound_expression),
+        )
+    
+    return resolve
+
+def cross(
+    left: Union[stp.Plan, UnboundPlan],
```
EpsilonPrime, line 220: Why does this take either a Plan or an UnboundPlan where other relations don't?  Is it worth adding a type definition for this Union so that all uses are kept in sync?

src/substrait/builders/type.py
```diff
@@ -0,0 +1,76 @@
...
+def struct(types: Iterable[stt.Type], nullable=True):
+    return stt.Type(struct=stt.Type.Struct(types=types, nullability=stt.Type.NULLABILITY_NULLABLE if nullable else stt.Type.NULLABILITY_REQUIRED))
+
+def list(type: stt.Type, nullable=True):
+    return stt.Type(list=stt.Type.List(type=type, nullability=stt.Type.NULLABILITY_NULLABLE if nullable else stt.Type.NULLABILITY_REQUIRED))
+
+def map(key: stt.Type, value: stt.Type, nullable=True):
+    return stt.Type(map=stt.Type.Map(key=key, value=value, nullability=stt.Type.NULLABILITY_NULLABLE if nullable else stt.Type.NULLABILITY_REQUIRED))
+
+def named_struct(names: Iterable[str], struct: stt.Type):
```
EpsilonPrime, line 75: This doesn't return an stt.Type so it probably doesn't belong in this file.  If you want to keep it here it needs a comment so that users don't mistakenly think that it is the same as others in this file.

src/substrait/builders/plan.py
tokoko: all of them should take in `Union[stp.Plan, UnboundPlan]`, I thought I made the changes everywhere, I'll fix it. type definition sounds like a good idea.

src/substrait/builders/plan.py
```diff
@@ -0,0 +1,294 @@
...
+                stp.PlanRel(
+                    root=stalg.RelRoot(input=rel, names=bound_inputs[0].relations[-1].root.names)
+                )
+            ],
+            **_merge_extensions(*bound_inputs),
+        )
+    
+    return resolve
+
+def fetch(plan: stp.Plan,
```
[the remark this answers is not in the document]
tokoko: yes, this is to allow passing along information about extensions and CTEs. Added a module-level comment.

src/substrait/builders/type.py
tokoko: I thought I could get away with it since `NamedStruct` is also defined in type_pb2. Do you think a return type annotation should be enough to show it's different?

PR #74, opened 2025-04-29
chore: switch to substrait antlr grammar

tokoko:
switches antlr generation to grammar definition from the main substrait repo.


src/substrait/gen/antlr/SubstraitType.interp
EpsilonPrime, line 1: This is a build artifact and can be deleted.  Same for tokens files

tokoko: done, thanks

EpsilonPrime (approved):

PR #81, opened 2025-06-06
feat: bindings for simple extensions

tokoko:
Adds python bindings for json schema of simple extensions.
Refactors extension registry to use new binding objects.


tokoko:
I chose to go with python dataclasses rather than something like pydantic as codegen target to minimize external dependencies. On the other hand, I had to also include a converter to convert parsed yaml into generated dataclass objects as standard library can't do that out of the box.

EpsilonPrime (approved):

PR #90, opened 2025-08-25
feat: pretty-printer for plan and expr

Clonkk:


Clonkk:
Let me know if you need further modification. The goal here is to have a good basis we can increment on to offer a nice debugging experience with substrait

Clonkk:
`utils.py` does sound nice 👍 I'll make the switch :) 

I wasn't aware of `substrait-textplan`, it seems like a super nice library. Indeed similar enough to what I'm doing.

I am just not a fan of having to load an external dependencies for something that's a debugging use-case essentially since it means i have to load a dependency I'm not using most of the time OR I have to add an external deps in my system every time I decide I need to add telemetry / logs to debug my substraits.


> If the goal there is to establish a standard text format for substrait

In my mind, the goal here is to have a nicer debugging experience when building systems around substraits. Substrait being protobuf the 'standard' should be [text_format](https://protobuf.readthedocs.io/en/latest/google/protobuf/text_format.html) I believe ?

mbrobbel (approved):

PR #97, opened 2025-10-08
fix: issue-96 call __str__ to get correct fn name

mbwhite:
The function name was being copied directly to the output structure, whereas it should have been created from the `__str__` on the function object.


src/substrait/builders/extended_expression.py
```diff
@@ -240,7 +240,7 @@ def resolve(
                 extension_function=ste.SimpleExtensionDeclaration.ExtensionFunction(
                     extension_uri_reference=registry.lookup_uri(uri),
                     function_anchor=func[0].anchor,
-                    name=function,
+                    name=str(func[0]),
```
tokoko, line 243: I think you should do the same for window_function and aggregate_function as well

mbwhite: well consistent at least.... the web pages don't spec what happen with no arguments 🤷

PR #104, opened 2025-10-16
fix: add nullability check for NamedStruct in builder

mbwhite:
Minor change, but useful to add defensive code where possible. 
The NamedStruct for a ReadRel shouldn't be nullable.  i.e. there must be some form of schema.

Adding in a simple check to the builders API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,5 +14,8 @@ codegen-extensions:
 lint:
 	uvx ruff@0.11.11 check
 
+lint_fix:
+	uvx ruff@0.11.11 check --fix
+
 format:
 	uvx ruff@0.11.11 format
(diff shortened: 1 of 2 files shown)


nielspardon (approved):
LGTM

for the `NamedStruct` type the struct should always have `NULLABILITY_REQUIRED` when used for schemas and `NamedStruct` is currently only used for either the base schema (ReadRel) or the table schema (WriteRel, DdlRel, UpdateRel) within the Substrait specification:

```proto
// * When used to represent a relation schema, the outermost struct's
//   nullability should be NULLABILITY_REQUIRED.
message NamedStruct {
  // list of names in dfs order
  repeated string names = 1;
  Type.Struct struct = 2;
}
```
https://github.com/substrait-io/substrait/blob/b958843bc5cc0df49e99456eea8c9f5391f66636/proto/substrait/type.proto#L291-L292

PR #105, opened 2025-10-17
feat: advanced_extensions

mbwhite:
Adding the ability to pass in an 'advanced extension' to the relation builders. 
Optional, but useful if needed. 

Improves the level of function compared to the other libraries


nielspardon:
@mbwhite can you rebase and resolve the conflicts? thanks

PR #106, opened 2025-10-17
chore: run tests with uv

tokoko:
switches to using uv lockfile to run tests to make ci environment more reproducible.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ cd substrait-python
 ```
 
 ## Development environment
-Activate environmen with uv.
+Activate environment with uv.
 ```
 uv sync --extra test
 ```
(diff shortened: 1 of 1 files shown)


nielspardon:
@tokoko what do you think about updating the https://github.com/substrait-io/substrait-python/blob/main/CONTRIBUTING.md file? that feels a bit out of sync to me if we are using `uv` I guess we don't need to tell people to use `conda` or am I missing something?

CONTRIBUTING.md
```diff
@@ -6,11 +6,10 @@ git clone --recursive https://github.com/<your-fork>/substrait-python.git
 cd substrait-python
 ```
 
-## Conda env
-Create a conda environment with developer dependencies.
+## Development environment
+Activate environmen with uv.
```
nielspardon: small typo

```suggestion
Activate environment with uv.
```

nielspardon:
it looks like the issue is caused by a protobuf version incompatibility. the `google.protobuf.any` introduced by @mbwhite seems to only exist in more recent versions of protobuf but not in version 3.20.1 which uv seems to pick currently.

the older version only has `google.protobuf.any_pb2`

PR #107, opened 2025-10-18
feat: allow passing multiple functions to function builders

tokoko:
changes builder api for all functions so one can easily pass multiple function refs to the builder.

### DIFF
--- a/src/substrait/builders/extended_expression.py
+++ b/src/substrait/builders/extended_expression.py
@@ -205,12 +205,12 @@ def column(field: Union[str, int], alias: Union[Iterable[str], str] = None):
 
 
 def scalar_function(
-    urn: str,
-    function: str,
+    function: Union[str, Iterable[str]],
     expressions: Iterable[ExtendedExpressionOrUnbound],
     alias: Union[Iterable[str], str] = None,
 ):
     """Builds a resolver for ExtendedExpression containing a ScalarFunction expression"""
+    functions = [function] if isinstance(function, str) else function
 
     def resolve(
         base_schema: stp.NamedStruct, registry: ExtensionRegistry
@@ -225,10 +225,16 @@ def scalar_function(
 
         signature = [typ for es in expression_schemas for typ in es.types]
 
-        func = registry.lookup_function(urn, function, signature)
+        for f in functions:
+            urn, name = f.rsplit(":", 1)
+            func = registry.lookup_function(urn, name, signature)
+            if func:
+                break
 
         if not func:
             raise Exception(f"Unknown function {function} for {signature}")
+
+        resolved_func, return_type = func
 
         func_extension_urns = [
             ste.SimpleExtensionURN(
@@ -288,7 +294,7 @@ def scalar_function(
                     ),
                     output_names=_alias_or_inferred(
                         alias,
-                        function,
+                        name,
                         [e.referred_expr[0].output_names[0] for e in bound_expressions],
                     ),
                 )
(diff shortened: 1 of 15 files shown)


tokoko:
- the builder will try looking up each function in the registry and throw an exception if none of them matches. this is practical because substrait often splits logically related functions in separate extensions, for example `["functions_arithmetic.yaml:add", "functions_arithmetic_decimal.yaml:add"]` while the user often doesn't care which one is used as long as the one with correct input types can be located.
- this should also make transition to urns a bit simpler as `uri` arg will no longer be part of the api.

it's a breaking change, but easy enough to change even if someone already actually depends on this stuff.

benbellick:
@nielspardon I felt that [this change in substrait-java](https://github.com/substrait-io/substrait-java/pull/564) was necessary specifically because it is indeed valid substrait to have multiple functions with the same implementation but different `urn`s. However, the calcite conversion caused an expected exception to be thrown if that was the case. Tt seemed like a questionable API to have valid substrait YAML files inadvertently cause a crash in a specific usage of the library. That being said, I am in no way wed to the resolution strategy I used there. I honestly chose it because it was simplest to implement, not because it was "best".

src/substrait/builders/extended_expression.py
```diff
@@ -205,12 +205,12 @@ def resolve(
 
 
 def scalar_function(
-    urn: str,
-    function: str,
+    function: Union[str, Iterable[str]],
```
benbellick: @tokoko I am open minded to this approach, but I would _strongly_ prefer not introducing string parsing if it is possible. Instead, maybe the API could be something like:
```suggestion
    function: Union[ExtensionID, Iterable[ExtensionID]],
```

where 
```python
@dataclass
class ExtensionID:
  urn: str
  function: str
```

What do you think?

benbellick (commented):
See comment

src/substrait/builders/extended_expression.py
benbellick: I would also be open minded to just having a separate function for the individual and the list case.

benbellick (commented):
Upon further thought, I am not sure I see the value of this. IIUC, what is happening is we find the first function with matching implementation in the list and use that one for extended_expression construction? 

To be this seems to be a lot less explicit than the current implementation. 

You write:
> while the user often doesn't care which one is used as long as the one with correct input types can be located.

Is that really true in general? The point of having different functions is to have different semantics after all, and I think the user should be expected to be as clear as possible which implementation they _intend_ on using, no? 

If we wanted to offer a facility for looking up the first matching function by name and signature irrespective of uri/urn, then I could imagine something like that in the `ExtensionRegistry`. 

If you feel strongly that this feature would be useful, then to me it makes more sense to exist as a separate helper function, rather than the canonical builder function. What do you think?

src/substrait/builders/extended_expression.py
tokoko: I'm not a fan of parsing strings either. I simply wanted not to clutter the api too much. How about using `NamedTuple` instead of a dataclass. It would allow the user to pass plain tuples as well.

```
from typing import NamedTuple, Union, Iterable

class ExtensionFunction(NamedTuple):
    urn: str
    function: str

def process_func(func: Union[ExtensionFunction, Iterable[ExtensionFunction]]):
    functions = [func] if isinstance(func[0], str) else func
    for f in functions:
        urn, name = f
        print(urn)
        print(name)

process_func(ExtensionFunction("sample_urn", "sample_func"))
process_func(("sample_urn", "sample_func"))
process_func([("sample_urn", "sample_func1"), ("sample_urn", "sample_func2")])
```

benbellick: I am okay with that approach as an API for allowing multiple parameters, though I still have my hesitations about the PR as a whole as expressed in the below comment.

PR #108, opened 2025-10-20
chore: fix and update devcontainer

nielspardon:
this PR fixes and updates the VS Code devcontainer:
- fix: the previous image `mcr.microsoft.com/vscode/devcontainers/python:3.9-buster` was dropped 
  by Microsoft so switching to a version without specifying the Debian base version and upgrading 
  to Python 3.13
- updates Graal to 25
- updates ANTLR to 4.13.2
- updates protoc to 30.2
- updates buf to 1.58.0
- updates uv to the latest version
- adds the --format=docker build option which makes the SHELL dockerfile instruction work with 
  podman
- adds a `postCreateCommand` that creates a virtualenv and install the dependencies into it
- adds some VS Code settings for pytest integration

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,13 +1,13 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.9-buster
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.13
 USER vscode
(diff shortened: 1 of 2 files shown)


tokoko (approved):
lgtm

andrew-coleman (approved):

PR #114, opened 2025-10-23
feat: graceful URI -> URN migration

### LINKED ISSUE #95: Migrate extensions from URIs to URNs
See https://github.com/substrait-io/substrait/issues/856 for context.

This will require temporarily handling both URIs and URNs, and then later dropping URI support.

This change depends on first bumping the version of substrait once a new substrait release occurs.

benbellick:
Introduces a graceful migration from the previous usage of URIs to URNs

- requires extensions to present a URI explicitly and a URN inside of the yaml file #95
- removes the usage of uri/urn aliases in the `ExtensionRegistry`
- introduce action to ensure examples run #121
- introduce action to ensure generated code up to date #120

Closes #95, #120, #121


tokoko:
Looks like the problem is in duckdb extension's json to substrait conversion, it is too strict and fails on unknown json fields. you can get around it by switching to use `from_substrait` instead of `from_substrait_json` in [test_sql_to_substrait.py](https://github.com/substrait-io/substrait-python/blob/711bd962ad76e46ab2f3c8628be6578b0d83f1b3/tests/sql/test_sql_to_substrait.py#L95):

```
sql = f"CALL from_substrait(?)"
substrait_out = conn.sql(sql, params=[plan.SerializeToString()])
```

that being said, we should probably open a ticket on duckdb extension as well to loosen json conversion checks.

tokoko:
@benbellick Can you introduce `make codegen-extensions` changes also?

src/substrait/builders/extended_expression.py
```diff
@@ -523,7 +575,7 @@ def resolve(
                 )
             ],
             base_schema=base_schema,
-            extension_uris=extension_uris,
```
tokoko: did you mean to remove this?

tokoko: this one looks good, but I think there are several others just like this in extended expression builders

benbellick: Ah got it. I was moving pretty fast to fix that one before having to run out. I'll be sure to look for the rest and push a commit. Thanks for catching these!

src/substrait/gen/json/simple_extensions.py
```diff
@@ -1,6 +1,6 @@
 # generated by datamodel-codegen:
 #   filename:  simple_extensions_schema.yaml
-#   timestamp: 2025-06-06T08:43:35+00:00
+#   timestamp: 2025-10-24T17:55:30+00:00
```
nielspardon: I guess we could disable adding the timestamp in `Makefile` by adding the `--disable-timestamp` flag to make repeated `make codegen-extensions` runs produce reproducible outputs

src/substrait/gen/proto/algebra_pb2.py
```diff
@@ -1,350 +1,347 @@
 """Generated protocol buffer code."""
+from google.protobuf.internal import builder as _builder
 from google.protobuf import descriptor as _descriptor
 from google.protobuf import descriptor_pool as _descriptor_pool
-from google.protobuf import runtime_version as _runtime_version
 from google.protobuf import symbol_database as _symbol_database
-from google.protobuf.internal import builder as _builder
-_runtime_version.ValidateProtobufRuntimeVersion(_runtime_version.Domain.PUBLIC, 5, 29, 5, '', 'proto/algebra.proto')
```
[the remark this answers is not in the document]
nielspardon: I guess we could add a Github Actions check that runs the codegen and tests whether the git workspace does not contain any uncommitted changes to help prevent missing to run codegen with the right versions

gen_proto.sh
```diff
@@ -18,7 +18,7 @@ rm -rf "$dest_dir/proto"
 
 # Generate the new python protobuf files
 buf generate
-protol --in-place --create-package --python-out "$dest_dir" buf
+uv run protol --in-place --create-package --python-out "$dest_dir" buf
```
benbellick: This was to make CI/CD not complain about missing `protol`

.github/workflows/codegen-check.yml
```diff
@@ -0,0 +1,44 @@
...
+            uv sync --extra test --extra gen_proto
+            # Run all code generation steps
+            make antlr
+            ./gen_proto.sh
+            make codegen-extensions
+
+      - name: Check for uncommitted changes (ignoring timestamps)
+        run: |
+          # Check for diffs, ignoring timestamp lines
+          if ! git diff -I '#.*timestamp:.*' --quiet --exit-code src/substrait/gen/; then
```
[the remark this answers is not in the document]
tokoko: Isn't `--disable-timestamp` flag that Niels suggested a cleaner fix for this? Did that not work for some reason?

PR #115, opened 2025-10-25
feat: configurable registry for sql conversion

tokoko:
Enables configuring extension registry in sql conversion


tokoko:
Just realized I was reloading extension files from disk on every test case in sql tests.

PR #125, opened 2025-11-15
feat: add select builder

tokoko:
The project builder in plan.py now builds a plan that outputs all inputs field
followed by all expressions.

The select builders captures the old behavior of project and outputs only the
expressions.

BREAKING CHANGE: the project builder in plan.py has been renamed to select


vbarua:
Fixes https://github.com/substrait-io/substrait-python/issues/124

PR #128, opened 2025-12-02
fix: correct if_then and switch_expression type inference calls

giospada:
Fixed type inference method calls in if_then and switch_expression builders.


nielspardon:
@giospada looks like you are reintroducing the substrait-cpp Git submodule which was previously removed from substrait-python. I don't think you need that module for this PR and could remove it.

PR #130, opened 2025-12-02
feat(types): add comprehensive type support and stricter validation

giospada:
Expanded type support with new scalar and parameterized types.
Improved parameter validation and stricter nullability/precision checks.


bestbeforetoday (commented):
Content looks good to me. Like tokoko, just one query/suggestion on code clarity (see inline comment). It's not a show-stopper though. If you want to change what is there, please go ahead. If not, just say and I think it could be merged as-is.

PR #131, opened 2025-12-03
feat: unify ANTLR setup and add pre_push make target

giospada:
- Create unified setup_antlr.sh script for ANTLR dependency
- Add setup-antlr make target for initialization
- Update Dockerfile to use unified setup script
- Add pre_push make target with all required checks
- Update CONTRIBUTING.md with pre-push instructions
- moved gen_proto.sh and update_proto.sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@ setup-antlr:
 	@bash scripts/setup_antlr.sh > /dev/null
 
 
-codegen: antlr codegen-proto codegen-extensions codegen-version
-
 
 antlr: setup-antlr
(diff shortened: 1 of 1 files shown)


tokoko:
@giospada this looks mergeable at the moment, but the thing is I'm thinking of changing the whole codegen setup to use pixi to avoid running the devcontainer in ci for codegen check. I checked and all the components we care about are available as conda packages (openjdk, libprotobuf, antlr, buf).

I can merge this as-is before that happens if it's causing you any headache though.

PR #135, opened 2025-12-24
chore: move dev extras to dependency-groups

tokoko:
Moves dev extras to dependency-groups


pyproject.toml
```diff
@@ -13,9 +13,12 @@ write_to = "src/substrait/_version.py"
 
 [project.optional-dependencies]
 extensions = ["antlr4-python3-runtime", "pyyaml"]
```
nielspardon: I'm wondering whether we still need this set of optional dependencies given that they are also part of the `dev` dependency group. Further, shouldn't those be default dependencies since we can assume most people with non-trivial use cases will want to use the extensions in some form?

nielspardon (approved):
LGTM

PR #136, opened 2026-01-05
feat: added FunctionType parameter to FunctionEntry

giospada:
- Added Function Type attribute to FunctionEntry
- Refactored extension registry on multiple files

### DIFF
--- a/tests/test_extension_registry.py
+++ b/tests/test_extension_registry.py
@@ -1,3 +1,5 @@
+import textwrap
+
 import pytest
 import yaml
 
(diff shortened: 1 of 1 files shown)


tests/test_extension_registry.py
```diff
@@ -631,3 +624,107 @@ def test_covers_struct_mismatched_types_fails():
...
+      - args: []
+        return: i64
+""",
+            "extension:test:window_funcs",
+            [("row_number", []), ("rank", [])],
+            "window",
+        ),
+    ]
+
+    for yaml_content, urn, functions, expected_type in test_cases:
```
tokoko: maybe use parametrized tests instead of running a loop inside a single test? (somewhat similar to [this](https://github.com/substrait-io/substrait-python/blob/799b8a991524b7979dc7380cde516af852b86b91/tests/sql/test_sql_to_substrait.py#L130)) results would be more user-friendly.

giospada: Yes i think is a good refactor

PR #137, opened 2026-01-05
chore: use pixi for codegen environment in ci

tokoko:


tokoko:
it's just a faster alternative to conda, written in rust of course :laughing: you can't configure system dependencies with uv (antlr, buf and protoc in this case). this PR only changes codegen check to use pixi which used whole devcontainer before. tests are still run with only uv. Although maybe it might make sense to port everything to pixi in the future as well for simplicity, not sure.

giospada:
Can you update the [CONTRIBUTING.md](https://github.com/substrait-io/substrait-python/blob/799b8a991524b7979dc7380cde516af852b86b91/CONTRIBUTING.md)?

Makefile
```diff
@@ -3,7 +3,7 @@ codegen: antlr codegen-proto codegen-extensions codegen-version
 
 antlr:
 	cd third_party/substrait/grammar \
-		&& java -jar ${ANTLR_JAR} -o ../../../src/substrait/gen/antlr -Dlanguage=Python3 SubstraitType.g4 \
+		&& antlr4 -o ../../../src/substrait/gen/antlr -Dlanguage=Python3 SubstraitType.g4 \
```
nielspardon: I cloned the PR branch and rebuild the devcontainer somehow it does not find the `antlr4` command when I run `make codegen` within the container.

.devcontainer/Dockerfile
```diff
@@ -1,13 +1,5 @@
...
-RUN curl -s "https://get.sdkman.io" | bash
-RUN bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && \
-    sdk install java 25-graalce"
-RUN mkdir -p ~/lib && cd ~/lib && curl -L -O http://www.antlr.org/download/antlr-4.13.2-complete.jar
-ENV ANTLR_JAR="~/lib/antlr-4.13.2-complete.jar"
-RUN cd ~ && curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v33.2/protoc-33.2-linux-x86_64.zip && \
-    unzip protoc-33.2-linux-x86_64.zip -d ~/.local && \
-    rm protoc-33.2-linux-x86_64.zip
-RUN curl -sSL "https://github.com/bufbuild/buf/releases/download/v1.58.0/buf-$(uname -s)-$(uname -m)" -o ~/.local/bin/buf && chmod +x ~/.local/bin/buf
 RUN curl -LsSf https://astral.sh/uv/install.sh | sh
```
nielspardon: Wouldn't pixi also take care of installing uv and we wouldn't need to install it separately? I guess the `postCreateCommand` in `devcontainer.json` may need to be changed then. And one would run the tests with `pixi run uv run pytest`.

nielspardon: I guess that would ensure that the uv version in the devcontainer is always consistent with the one defined in pixi since theoretically the devcontainer might one day start installing a uv version that is outside of the dependency range defined in the pixi definitions.

tokoko: I couldn't get activation to work properly yet. vscode seems to activate both pixi and uv multiple times and clutters up stdout. I'll look into that in the future, but for now I went the opposite direction. I moved pypi dependencies that we were using for codegen to pixi side (and removed uv from there). Now `pixi run make codegen` no longer depends on uv.

tokoko: done

Makefile
```diff
@@ -13,16 +12,17 @@ codegen-version:
...
 		--input third_party/substrait/text/simple_extensions_schema.yaml \
 		--output src/substrait/gen/json/simple_extensions.py \
 		--output-model-type dataclasses.dataclass \
 		--target-python-version 3.10  \
-		--disable-timestamp
+		--disable-timestamp \
+		--formatters ruff-format
 
 lint:
 	uvx ruff@0.11.11 check
```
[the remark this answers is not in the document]
tokoko: moved everything from make to pixi

check_codegen.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+if ! git diff --quiet --exit-code src/substrait/gen/; then
+    echo "Code generation produced changes. Generated code is out of sync!"
+    echo ""
+    git diff src/substrait/gen/
+    echo ""
+    echo "To fix this, run:"
+    echo "  make codegen"
```
nielspardon:
```suggestion
    echo "  pixi run codegen"
```

pyproject.toml
```diff
@@ -34,3 +32,58 @@ respect-gitignore = true
...
+dev = { features = ["dev"], solve-group = "default" }
+extensions = { features = ["extensions"], solve-group = "default" }
+sql = { features = ["sql"], solve-group = "default" }
+
+[tool.pixi.tasks]
+format = "ruff format"
+lint = "ruff check"
+lint_fix = "ruff check --fix"
+
+update-submodule = "./update_submodule.sh"
```
nielspardon: it feels to me for the time being (until https://github.com/substrait-io/substrait/issues/925) one would always run codegen after updating the submodule. what do you think about linking them?

nielspardon: aka update-submodule triggers codegen but you can still do codegen independently

PR #145, opened 2026-01-14
feat: remove generated protobuf code from substrait-python

vbarua:
Generation of protobuf bindings has been extracted to the substrait-protobuf package.

BREAKING CHANGE: code under `substrait.gen.proto` is now found under `substrait`
BREAKING CHANGE: simple extension files have been moved from `extensions` to `extension_files`
BREAKING CHANGE: version information has been moved to `substrait.version` module


src/substrait/proto/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+import substrait.type_pb2 as type
+from substrait.algebra_pb2 import *
+from substrait.capabilities_pb2 import *
+from substrait.extended_expression_pb2 import *
+from substrait.extensions.extensions_pb2 import *
+from substrait.function_pb2 import *
+from substrait.parameterized_types_pb2 import *
+from substrait.plan_pb2 import *
+from substrait.type_expressions_pb2 import *
+from substrait.type_pb2 import *
```
vbarua: I'm not clever enough to recreate whatever was happening in `proto.py`, but I am smart enough to re-export it all by hand. It's not as flashy, but it does the job.

src/substrait/__init__.py
```diff
@@ -1,8 +0,0 @@
-try:
-    from ._version import __version__  # noqa: F401
-except ImportError:
-    pass
-
-__substrait_version__ = "0.79.0"
-__substrait_hash__ = "92d2e75"
-__minimum_substrait_version__ = "0.30.0"
```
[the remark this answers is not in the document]
tokoko: this made me think if we might have a problem if someone inadvertently installs different versions of `substrait-protobuf` and `substrait-antlr` for example... do you think `substrait_version` should refer exclusively to protobuf? or maybe we need some sort of an eager check to make sure all those substrait package versions line up

src/substrait/proto/__init__.py
tokoko: tbh, pretty sure that module was there to avoid writing out `gen.proto` in imports (which I was doing anyway :laughing:) if protos are moved to top-level `substrait` namespace, I don't think we need to keep this around.

vbarua: So without this we actually get test failures like
```
AttributeError: module 'substrait.proto' has no attribute 'ExtendedExpression
```
in https://github.com/substrait-io/substrait-python/blob/main/tests/sql/test_sql_to_substrait.py, which seem to be coming from the following line in [pyarrow](https://github.com/apache/arrow/blob/9cde706419c8e9cde0647af1fd55f6a86ecdbd2a/python/pyarrow/_substrait.pyx#L34):
```python
    import substrait.proto  # no-cython-lint
```

We should probably update the code there to not rely on that import, but I think we want to do that _after_ we have the new path available in this library.

tokoko (approved):
looks good

nielspardon (approved):

PR #157, opened 2026-03-13
build: enable Trusted Publishing to PyPI

vbarua:


vbarua:
In the process of working on substrait-packaging, I learnt about PyPI [trusted publishing](https://github.com/pypa/gh-action-pypi-publish#trusted-publishing).

I've configured the publisher in PyPI associated with the `release.yml` workflow
Screenshot 2026-03-13 at 12 36 39

Using this, we can avoid having to persists a long-lived token, and it gives users a better guarantee that the package they are installing was generated from a CI job.

tokoko (approved):
nice

PR #168, opened 2026-06-03
feat: bump substrait packages to 0.86.0 and support additional extension metadata

nielspardon:
Bump `substrait-protobuf`, `substrait-extensions`, and `substrait-antlr` from 0.85.0 to 0.86.0.

Add support for the new 0.86.0 simple-extension schema fields when parsing extension YAML:

- deprecation status (`since`/`reason`/`metadata`) on types, type variations, functions, and
  individual implementations
- implementation-level `description`
- function- and type-level `metadata`

The 0.86.0 spec changes are otherwise contained in the vendored extension YAML shipped by the
packages (no proto changes), so no further code modifications were required.

All 231 tests pass (3 pre-existing xfails, unchanged from the 0.85.0 baseline).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,8 +7,8 @@ readme = "README.md"
 requires-python = ">=3.10,<3.15"
 dependencies = [
     "protobuf >=5,<7",
-    "substrait-protobuf==0.85.0",
-    "substrait-extensions==0.85.0",
+    "substrait-protobuf==0.86.0",
+    "substrait-extensions==0.86.0",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -7,20 +7,20 @@ readme = "README.md"
...
+    "substrait-extensions==0.86.0",
 ]
 dynamic = ["version"]
 
 [tool.setuptools_scm]
 write_to = "src/substrait/_version.py"
 
 [project.optional-dependencies]
-extensions = ["antlr4-python3-runtime", "substrait-antlr==0.85.0", "pyyaml"]
+extensions = ["antlr4-python3-runtime", "substrait-antlr==0.86.0", "pyyaml"]
```
tokoko: do we still need `antlr4-python3-runtime` here? transitive dependency should be enough.

nielspardon: good point. we can clean that up. I pushed a commit that removes it and updates uv.lock

tokoko (approved):
lgtm

PR #171, opened 2026-06-08
ci: automate releases with semantic-release

nielspardon:
## What

Automate releases with
[semantic-release](https://semantic-release.gitbook.io/), matching the
model used by other Substrait repos such as substrait-java.

## How it fits together

A new scheduled (weekly) + manual `Semantic Release` workflow runs
semantic-release under a GitHub App token. It analyzes Conventional
Commits, computes the next `vX.Y.Z`, updates `CHANGELOG.md`, pushes the
tag, and creates a GitHub Release with generated notes.

The existing `release.yml` keeps its filename (the PyPI trusted-publisher
binding depends on it) and is repurposed to trigger on that tag: it
builds the package, attaches the wheel/sdist to the release that
semantic-release created, and publishes to PyPI via Trusted Publishing.

This is a hybrid of the java model rather than a literal copy, because
substrait-python derives its version from the git tag via setuptools_scm
(no version file to bump) and publishes via Trusted Publishing / OIDC,
which must run in `pypa/gh-action-pypi-publish` rather than a shell
`publishCmd`.

## Changes

- New `ci/release/run.sh` — runs `semantic-release --ci` via npx with a
  pinned toolchain (a floating conventional-changelog preset silently
  dropped the Features/Bug Fixes sections from the notes).
- New `.releaserc.mjs` — conventionalcommits preset on main, breaking to
  minor (pre-1.0, matching java), changelog + github + git plugins. An
  ESM config so the side-effecting github/git plugins can be gated behind
  `RELEASE_DRY_RUN=false` (allowing a credential-free dry run) and so the
  release-notes-generator can strip git trailers (`Signed-off-by`, etc.)
  the parser would otherwise fold into BREAKING CHANGE notes.
- New `ci/release/dry_run.sh` + `.github/workflows/release-notes-check.yml`
  — a credential-free dry run against the real history (in a throwaway
  worktree) that fails if a release is due but its notes come out with no
  sections, guarding against a toolchain regression.
- New `.github/workflows/semantic-release.yml` — scheduled + manual,
  runs under a GitHub App token so the pushed tag triggers release.yml.
- Changed `.github/workflows/release.yml` — tags-only trigger; a
  `github-release` job waits for the semantic-release release to appear
  and attaches the wheel/sdist to it (leaving the generated notes
  intact), and PyPI publish depends on that job.
- Changed `RELEASING.md` — documents the automated flow and the
  release-notes guard.

## Required before this works

1. The RELEASER GitHub App (RELEASER_ID / RELEASER_KEY secrets) must be
   installed on this repo with push and branch-protection-bypass rights.
2. The PyPI Trusted Publisher must remain bound to workflow release.yml
   and environment pypi (preserved by this PR).

## Verification done

- shellcheck on `ci/release/run.sh` and `ci/release/dry_run.sh` — clean
- MJS / YAML / bash syntax validated
- Ran `ci/release/dry_run.sh` against real history (last tag `v0.29.0` →
  `0.30.0`): the notes render the expected Features / Bug Fixes sections
- Verified git-trailer stripping end-to-end: a BREAKING CHANGE commit
  with `Signed-off-by` / `Co-authored-by` trailers produces clean notes

Closes https://github.com/substrait-io/substrait-python/issues/147


nielspardon:
I did create a new `substrait-python releaser` Github App on the substrait-io organization and configured GH Actions secrets for `RELEASER_ID` and `RELEASER_KEY` for the substrait-python repo.

.github/workflows/release.yml
```diff
@@ -33,27 +34,10 @@ jobs:
...
-        uses: softprops/action-gh-release@v3
-        with:
-          files: |
-            ./dist/*.whl
-            ./dist/*.tar.gz
   publish:
     name: Publish to PyPI
     runs-on: ubuntu-latest
-    needs: release
+    needs: build
```
bvolpato: this drops GitHub release assets. `v0.29.0` and previous releases include wheel and sdist files, but semantic-release creates release before this tag workflow runs and nothing now attaches `dist/`. Please keep asset-upload job (it can attach to existing release) and make PyPI publish depend on it.

ci/release/run.sh
```diff
@@ -0,0 +1,14 @@
...
+set -euo pipefail
+
+npx --yes \
+  -p semantic-release \
+  -p "@semantic-release/commit-analyzer" \
+  -p "@semantic-release/release-notes-generator" \
+  -p "@semantic-release/changelog" \
+  -p "@semantic-release/github" \
+  -p "@semantic-release/git" \
+  -p "conventional-changelog-conventionalcommits" \
```
bvolpato: floating versions already break note generation. I ran an isolated dry run against this head with this config: 27 commits, next version `0.30.0`, but output contained only the release heading despite two `feat` and two `fix` commits. Pinning `conventional-changelog-conventionalcommits@9.3.1` restored `Features` and `Bug Fixes`. Please pin release toolchain and add a credential-free dry-run check that asserts expected sections.

nielspardon: Thanks for catching that. I also found that issue on substrait-java recently and tried to fix it in https://github.com/substrait-io/substrait-java/pull/989. I will apply the same fix in this PR.

.github/workflows/release.yml
nielspardon: Good catch. Let me fix that.

nielspardon: Done in f052491. Restored an asset-upload job in `release.yml`: it waits for the release semantic-release creates for the tag to appear (polling `gh release view`), then attaches the wheel/sdist to that existing release — omitting `body` so the generated notes are preserved. PyPI `publish` now `needs` that job.

ci/release/run.sh
nielspardon: Done in f052491. Pinned the whole toolchain (including `conventional-changelog-conventionalcommits@9.3.1`) in `ci/release/run.sh`, and moved the plugin set to `.releaserc.mjs` so the side-effecting github/git plugins are omitted unless `RELEASE_DRY_RUN=false` — that lets the dry run run without credentials. Added `ci/release/dry_run.sh` plus a `Release Notes Check` workflow that runs the pinned toolchain against synthetic feat/fix history and fails if the `Features`/`Bug Fixes` sections are missing.

andrew-coleman (approved):

PR #172, opened 2026-06-08
ci: remove PR title check comment when check passes

nielspardon:
Ports https://github.com/substrait-io/substrait-java/pull/850 to substrait-python.

Previously the PR Title Check workflow posted an "ACTION NEEDED" comment when the PR
title/description did not follow Conventional Commits, but never removed it once the title was
fixed.

This consolidates the comment logic into a single `github-script` step that branches on the
commitlint outcome (using `continue-on-error: true` plus `core.setFailed(...)` to keep the check
red on failure):

- On pass: delete any previously posted comment.
- On fail: post the comment if not already present.

The bot comment is identified via a hidden `<!-- pr-title-check -->` marker rather than exact body
matching, so cleanup survives wording or whitespace changes. The workflow now declares
`permissions: pull-requests: write` so it can delete comments.

The Python-specific `.commitlintrc.js` content (the dependabot `Bumps [...]` ignore rule) is
preserved.

Note: cleanup is marker-based, so it only applies to comments posted by this new version going
forward. Pre-existing comments from the old workflow version won't be automatically removed.

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -3,6 +3,10 @@ name: PR Title Check
 on:
   pull_request_target:
     types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
 jobs:
   commitlint:
     name: PR title / description conforms to semantic-release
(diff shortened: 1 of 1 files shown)


tokoko:
looks good, but just wondering.. what's the point of the comment in the first place? isn't red check in checks section enough?

nielspardon:
> looks good, but just wondering.. what's the point of the comment in the first place? isn't red check in checks section enough?

We have this across most repositories and I'm only trying to improve it for now. I think the message is meant to give a slightly better explanation than the commitlint failure messages of the check.

bestbeforetoday (approved):
Seems like a good improvement to what is there now. We can think about whether the comment is necessary and, if not, address it in a future change.

PR #178, opened 2026-06-16
fix: handle unparameterized interval_day type

### LINKED ISSUE #177: interval_day type without precision parameter causes error in derivation expression evaluation
When using interval_day without an explicit precision parameter (e.g., in an extension function's return type derivation), the evaluation fails with an error. The _evaluate function in derivation_expression.py assumes the precision parameter is always present for PrecisionIntervalDayContext, but the Substrait spec allows interval_day to be used without specifying precision.

## Steps to Reproduce

Use interval_day (or interval_day?) as a type expression without a precision parameter:

from substrait.derivation_expression import evaluate

evaluate("interval_day")  # raises an error

### Expected Behavior

interval_day without a precision parameter should evaluate to a valid Type.IntervalDay message (with default/unset precision), similar to how other types handle optional parameters.

### Actual Behavior

The evaluation attempts to call _evaluate on None (the missing precision), resulting in an error.

thgschwind:
Allow interval_day to be used without an explicit precision parameter.

Fixes: #177

### DIFF
--- a/src/substrait/derivation_expression.py
+++ b/src/substrait/derivation_expression.py
@@ -151,6 +151,12 @@ def _evaluate(x, values: dict):
             elif isinstance(
                 parametrized_type, SubstraitTypeParser.PrecisionIntervalDayContext
             ):
+                if parametrized_type.precision is None:
+                    return Type(
+                        interval_day=Type.IntervalDay(
+                            nullability=nullability,
+                        )
+                    )
                 precision = _evaluate(parametrized_type.precision, values)
                 return Type(
                     interval_day=Type.IntervalDay(
--- a/tests/test_derivation_expression.py
+++ b/tests/test_derivation_expression.py
@@ -129,6 +129,15 @@ def test_interval_day():
     )
 
 
+def test_interval_day_no_precision():
+    assert evaluate("interval_day") == Type(
+        interval_day=Type.IntervalDay(nullability=Type.NULLABILITY_REQUIRED)
+    )
+    assert evaluate("interval_day?") == Type(
+        interval_day=Type.IntervalDay(nullability=Type.NULLABILITY_NULLABLE)
+    )
+
+
 def test_struct_simple():
     """Test simple struct with two i32 fields."""
     result = evaluate("struct<i32, i32>", {})


src/substrait/derivation_expression.py
```diff
@@ -151,6 +151,12 @@ def _evaluate(x, values: dict):
             elif isinstance(
                 parametrized_type, SubstraitTypeParser.PrecisionIntervalDayContext
             ):
+                if parametrized_type.precision is None:
+                    return Type(
+                        interval_day=Type.IntervalDay(
+                            nullability=nullability,
+                        )
+                    )
```
nielspardon, line 159: Do you have a little bit more context on whether this is only an issue for `IntervalDay` for you and how this is triggered? I'm asking since we have very similar code above for `PrecisionTimestampContext` and `PrecisionTimestampTZContext`.

nielspardon:
Thanks for putting this together, @thgschwind, and apologies for the slow turnaround.

Revisiting this in light of the [substrait v0.97.0 spec change](https://github.com/substrait-io/substrait/releases/tag/v0.97.0) (July 12): the premise that `interval_day` may be used without a precision no longer holds, so I don't think we can take this approach.

- **Proto:** v0.97.0 changed `IntervalDay.precision`'s contract to *"Implementations should reject an unset precision"* (previously an unset precision was treated as `6`/microseconds). This PR does the opposite — it *produces* a `Type.IntervalDay` with **unset** precision, i.e. exactly the message the spec now says to reject.
- **Grammar:** `interval_day` has no bare form in `SubstraitType.g4` — the only alternative is `interval_day<precision>`. `evaluate("interval_day")` is actually a syntax error (`mismatched input '<EOF>' expecting {'<','?'}`); it only reaches the `precision is None` branch because ANTLR error recovery hands back a partial parse tree.

Revisiting my earlier question about whether this is IntervalDay-specific — it isn't. Every precision-parameterized type (`decimal`, `interval_day`, `interval_compound`, `precision_time`, `precision_timestamp`, `precision_timestamp_tz`) requires explicit parameters, so bare forms are all invalid in the same way.

Closing without merging, since defaulting to unset precision is now spec-noncompliant. The one real remaining problem — the confusing `Unknown token type NoneType` on invalid syntax — is worth fixing on its own and I've split it out into #216.

PR #195, opened 2026-06-17
build: restrict permissions for commitlint

bestbeforetoday:
The pr-title workflow gives elevated pull-request write permissions to the third-party commitlint tool. This is a potential supply-chain vulnerability.

This change restricts the permissions to only read for the commitlint tool.

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -4,13 +4,16 @@ on:
   pull_request_target:
     types: [opened, edited, synchronize, reopened]
 
+# Default to the minimum read-only token for all jobs.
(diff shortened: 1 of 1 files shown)


tokoko (approved):
tbh I'd rather we don't post the PR message at all 😆 but this is better than nothing

bestbeforetoday:
> tbh I'd rather we don't post the PR message at all 😆 but this is better than nothing

I am also not overly keen on the way comments are added and then removed. It would be nice if there was some other way to provide supporting information in the **Details** link when the PR check workflow fails. The DCO sign-off check app does this, but I don't know how.

PR #204, opened 2026-07-02
feat: ergonomic native DataFrame/Expr API (substrait.api)

nielspardon:
## Summary

substrait-python can already build any plan, but the day-to-day building experience lags behind sibling libraries — notably [substrait-java](https://github.com/substrait-io/substrait-java), whose `core` module ships a fluent builder, a nullability-aware type factory, a preloaded extension collection, and named function helpers.

This PR adds a **thin, additive facade** over the existing `substrait.builders` layer that gives Python users the idioms they expect from pandas / Polars / PySpark / Ibis:

```python
import substrait.dataframe as sub

plan = (
    sub.read_named_table("people", {"id": sub.i64, "age": sub.i64, "name": sub.string})
    .filter(sub.col("age") > 25)
    .with_columns(bonus=(sub.col("age") + 1) * 2)
    .select("id", "name", "bonus")
    .to_plan()
)
```

The facade is **faithful**: each verb / expression emits byte-identical protobuf to the equivalent raw builder call (asserted per feature). It began as `filter`/`select`/`join` and has since grown to cover essentially the whole Substrait query-building surface, and moves the project onto the latest spec (0.96).

> Draft — opening for feedback on the API shape, the spec bump, and the Narwhals direction.

## What's here

**Entry point** — `import substrait.dataframe as sub`. The higher-level API lives under a dedicated `substrait.dataframe` subpackage rather than at the `substrait` namespace root: `substrait` is a PEP 420 namespace package shared with `substrait-protobuf` (so the root can't hold an `__init__.py`), and grouping the modules under one owned subpackage — a sibling to `builders` / `sql` / `narwhals`, named for what it is rather than the generic "api" — keeps a single import surface and avoids scattering generic names (`expr`, `frame`, …) across the shared namespace. Nothing in `builders` / `proto` behaviour changes.

**Expressions (`Expr`)** — operator overloading (`< <= > >= == != + - * / % ** & | ^ ~`) mapping to the standard function extensions and resolving lazily, including decimal arithmetic (`functions_arithmetic_decimal`) for decimal operands; literal auto-wrap with peer-type coercion (`col_fp64 * 2`, `col_i32 > 25`, `col_decimal > Decimal("9.99")`), which raises rather than silently rounding a literal the column's decimal type can't represent exactly. Plus `cast`, `alias`, `is_null` / `is_not_null` / `is_nan`, `between`, `is_distinct_from`, `is_in`, `coalesce`, `when().then()…otherwise()` and `switch` (CASE), nested access (`struct_field` / `[]` / `map_key`), window functions via `.over(partition_by=, order_by=, rows=/range=)`, higher-order list functions (`list_transform` / `list_filter` with Python-lambda callbacks), subqueries (`scalar_subquery` / `exists` / `unique` / `in_subquery` / `any_` / `all_`, including **correlated** via `outer()`), dynamic parameters (`parameter`), and execution-context variables (`current_timestamp` / `current_date` / `current_timezone`).

**DataFrame verbs**
- reads: `read_named_table`, `from_records` (VALUES), `read_parquet` / `read_csv` / `read_orc` / `read_arrow`, `read_extension_table`, `extension_leaf`
- projection/rows: `filter`, `select`, `with_columns`, `rename`, `drop`, `sort` (per-column direction / nulls), `limit` / `head` / `offset`, `top_n`
- joins: `join` (all 13 join types + `post_filter`), `cross_join`, `nested_loop_join`, `hash_join`, `merge_join`
- set ops: `union` / `intersect` / `except_`
- aggregation: `group_by().agg()`, `rollup`, `cube`, explicit grouping sets, per-measure `FILTER`, `DISTINCT`, ordered aggregates
- reshape/physical: `unpivot`, `repartition`, `broadcast`, `extension` / `extension_multi`
- plan structure: `hint`
- write / DDL: `write_named_table` (CTAS / insert), `create_table` / `create_view`, `drop_table` / `drop_view`, `update_table`

**Function namespace (`f`)** — every scalar / aggregate / window function from the default extensions, generated lazily from the registry (`dir(sub.f)`-discoverable). Multi-extension names resolve to the right extension by argument type; keyword names via `and_` / `or_` / `not_`; function options as kwargs (`f.add(x, y, overflow="ERROR")`). `functions_for(registry)` / `DataFrame.f` surface custom-extension functions.

**Types** — nullability-aware shortcuts (`sub.i64` nullable, `sub.i64.non_null` required) covering every concrete Substrait type; parametrized builders; `user_defined` UDTs.

**User-defined extension relations** — `ExtensionLeafDetail` / `ExtensionSingleDetail` / `ExtensionMultiDetail` ABCs (mirroring substrait-java's `Extension.*RelDetail`) with `to_any` / `from_any` / `derive_schema`, registered via `ExtensionRegistry.register_extension_relation` so schema inference follows a custom relation like a built-in one.

## Spec bump: 0.86 → 0.96

Moves the `substrait-protobuf` / `substrait-extensions` / `substrait-antlr` pins to the latest spec (needed for `TopNRel` and execution-context variables, and to track upstream). Done as a **bisected, two-step migration**:
- **removed spec features absorbed** — `AggregateRel.Grouping.grouping_expressions` and the deprecated `timestamp` / `time` / `timestamp_tz` types (superseded by `expression_references` / the `precision_*` types), across `type_inference`, `derivation_expression`, and tests;
- **ANTLR type-derivation grammar refactor** (0.95 split the single `BinaryExpr` rule into `MulDiv` / `AddSub` / `Comparison` / `Equality` / `And` / `Or` + `If` / `Not`) migrated in `derivation_expression`.

## Core enablers (beyond the pure facade)

Most verbs are thin wrappers, but a few features required **additive** changes to the lower layers — never a reimplementation; existing paths stay byte-identical:
- `type_inference` gained schema-inference cases for the new relations (`expand`, `top_n`, `exchange`, physical joins, extension relations) and expression types (`lambda` → `func<>`, `dynamic_parameter`, `execution_context_variable`), plus outer-schema resolution for correlated references;
- the registry's signature matcher gained `func<…>` matching so higher-order functions resolve and bind type parameters;
- correlated subqueries thread a small contextvar (an outer-schema stack) so inference resolves outer references.

## Breaking changes

- **Spec 0.86 → 0.96**: the deprecated `timestamp` / `time` / `timestamp_tz` types and `AggregateRel.Grouping.grouping_expressions` are gone (superseded by the `precision_*` types / `expression_references`).
- **`substrait.dataframe` reclaimed** — the old (experimental, minimal) Narwhals wrapper that lived at `substrait.dataframe` moves to `substrait.narwhals`, and `substrait.dataframe` **now is** the native DataFrame/Expr API (the entry point earlier drafted as `substrait.api`). Consequence: existing `import substrait.dataframe` code silently binds to the native API instead of the Narwhals wrapper (a behaviour change, not an `ImportError`) — migrate those imports to `substrait.narwhals`.

## Testing

New `tests/dataframe/` plus builder golden tests, all asserting byte-identical protobuf to the raw builder path, with schema-inference and round-trip checks. Engine round-trip tests (pyarrow / DuckDB / DataFusion) lag the spec and can crash the interpreter natively on a newer-than-supported plan, so per a best-effort policy they are **skipped unless `SUBSTRAIT_ENGINE_TESTS=1`**; the non-engine suite is the gate and is green.

## Follow-ups (not in this PR)

- CTEs / shared subplans (`ReferenceRel`) at both the builder and dataframe layers (`DataFrame.cache()`), carrying subtrees in-band in the `Plan` rather than via a contextvar — tracked in #211 (prototyped here, then pulled out to keep this PR focused).
- Build `substrait.narwhals` out into a full Narwhals *compliant* backend (CompliantLazyFrame / Expr / Namespace) on top of `substrait.dataframe.frame`. The compliant protocol is experimental, and `collect()` implies execution (a non-goal), so it would raise or delegate to a consumer.
- Positional enum function arguments (`FunctionArgument.enum`); type variations on built-in types; Iceberg read specifics (`ReadRel.IcebergTable`).
- Registry-derived operator resolution: resolve the arithmetic/comparison operators (`+`, `>`, …) over *every* extension the runtime registry defines for the function name — via the shared `_resolve_over_urns` loop, replacing the hardcoded standard + `_decimal` URN list — so operators work over user-defined types and type variations from custom extension YAMLs (the peer literal-coercion step, the operators' reason to differ from `f.*`, barely applies to UDTs). Broadens the overloads operators consider (e.g. `date + interval` via `functions_datetime`), so it needs its own tests and an intent decision.

🤖 Generated with AI


tokoko:
I haven't looked through everything yet, but my +1 on the overall approach. There are some cases (lambda functions most notably) where narwhals/polars api and substrait representation genuinely diverge from one another. Having a native api and a narwhals wrapper is probably a better long-term decision.

nielspardon:
> I haven't looked through everything yet, but my +1 on the overall approach. There are some cases (lambda functions most notably) where narwhals/polars api and substrait representation genuinely diverge from one another. Having a native api and a narwhals wrapper is probably a better long-term decision.

Yeah, definitely a lot of Substrait features to be covered. Wasn't sure yet how to stage it and thought I stop after the first couple iterations to see what the community thinks.

mbwhite:
Also +1 for the general approach.  

I think there was a discussion on a Dask issue about DataFrames/Substrait... will need to find it

EDIT https://github.com/dask-contrib/dask-sql/pull/1041

nielspardon:
I know this is a massive PR. I tried creating scoped commits. Let me know if you prefer to spread this out over multiple PRs.

mbwhite:
On the schema that's supplied eg `read_named_table` I wonder if that should take a wrapper - (maybe even substrait's `named_struct`?) – and to have helpers to move between common representations, such as Arrow.

Based on the assumption that a user would be coming from a situation where they may already have a schema. They look at the API  "oh I need to convert the schema again..."

nielspardon:
> On the schema that's supplied eg `read_named_table` I wonder if that should take a wrapper - (maybe even substrait's `named_struct`?) – and to have helpers to move between common representations, such as Arrow.
> 
> Based on the assumption that a user would be coming from a situation where they may already have a schema. They look at the API "oh I need to convert the schema again..."

might be best to log this as a follow-up issue to not make this PR even bigger. mapping other representations like e.g. narwhals into the Substrait native dataframe is also already a follow-up activity.

tokoko:
@mbwhite somewhat related there's a `read_adbc_named_table` function in this [example](https://github.com/substrait-io/substrait-python/blob/main/examples/adbc_example.py) which I was tempted to include somehow in the library, but couldn't really find a good home for it.

@nielspardon looks good so far, I'll try to take a thorough look today or tomorrow.

src/substrait/dataframe/__init__.py
```diff
@@ -0,0 +1,173 @@
+"""Ergonomic front door for substrait-python.
```
tokoko, line 1: curious, is there a reason this needs to be a file rather than a folder? I'm also not sure about naming. The way I thought about it before both `builders` and `dataframe` were an api, one lower-level and another higher-level.

Did you decide on `api` name to signal that this one is the preferred entrypoint for users?

nielspardon: was not my decision but Claude Codes when I told it to compare what we have in substrait-java with substrait-python and it saw the `SubstraitBuilder` class in substrait-java

I don't have a strong opinion on that naming and happy to change it.

nielspardon: I remember it considered putting it at the `substrait` module level but that did not work since the substrait module is shared across substrait-python, substrait-protobuf, substrait-antlr, substrait-extensions so it went for `api.py` as the entry point.

tokoko: tbh, I'd prefer to avoid polluting substrat module level namespace. how about something like this?

```
api/
  __init__.py
  dtypes.py
  expr.py
  extension_relations.py
  frame.py
  functions.py
```

`substrait.expr` with become `substrait.api.expr` and so on, just to group all higher-level api code under a single module.

In terms of naming, I'm fine with both `substrait.api` and `substrait.dataframe` with slight preference for `substrait.dataframe` because I think that's more descriptive, plus it's not the only "api". `builders` is lower-level api, `sql` and `dataframe` can both be different higher-level apis.

nielspardon: Done — went with your suggestion. The higher-level modules now live under a single `substrait.dataframe` subpackage instead of scattered at the shared namespace root:

```
substrait/dataframe/
  __init__.py            # the facade (was substrait/api.py)
  dtypes.py
  expr.py
  extension_relations.py
  frame.py
  functions.py
```

So `substrait.expr` → `substrait.dataframe.expr` and friends, and the entry point is now `import substrait.dataframe as sub`. Renamed `tests/api/` → `tests/dataframe/` and the example to match. Pushed in `cbfdc4c` (+ an `import`-sort follow-up in `10f70c2`), and I've updated the PR description to match.

On the name: agreed on `substrait.dataframe` over `api` — it sits alongside `builders` (lower-level) / `sql` / `narwhals` as a sibling named for what it is, rather than claiming the generic "api". One consequence I called out in the breaking-changes section: since this PR already moves the old experimental Narwhals wrapper `substrait.dataframe` → `substrait.narwhals`, the name `substrait.dataframe` is now *reclaimed* for the native API — so existing `import substrait.dataframe` code silently binds to the native API instead of the wrapper (a behaviour change, not an `ImportError`).

One question back to you: the import alias. I kept `sub` (`import substrait.dataframe as sub`) — it reads well at call sites (`sub.col`, `sub.i64`, `sub.f.sum`) and is mnemonic for the brand, though it's a touch odd on the import line itself. `df` is out (conventionally a frame *instance*). Do you have a preference, or is `sub` fine?

tokoko: `sub` sounds good. I'd not worry about the previous dataframe impl, that was just a stub.

src/substrait/dataframe/expr.py
```diff
@@ -0,0 +1,914 @@
...
+    def __ne__(self, other: Any) -> "Expr":  # type: ignore[override]
+        return self._compare(other, "not_equal")
+
+    # Operator-overloaded ``__eq__`` means an Expr is not a normal value; like
+    # pandas/Polars expressions it is intentionally not hashable.
+    __hash__ = None  # type: ignore[assignment]
+
+    # -- arithmetic -------------------------------------------------------
+    def __add__(self, other: Any) -> "Expr":
+        return _numeric_binary(self, other, FUNCTIONS_ARITHMETIC, "add")
```
tokoko: this won't work for decimals, right? any reason why we can't use multi urn helper here as well?

nielspardon: yeah, I also had noticed that. I don't think there is a good reason. Let me request this to be added.

nielspardon: this does support decimals with operators now. I did add a follow-up to the PR description for adding user defined type / type variation support as well to operators for advanced users

mbwhite:
I've found one 'gotcha' ... this API is creating the FetchRel (https://github.com/substrait-io/substrait/blob/858397de179721218bb5ca4a329121ee053e9cab/proto/substrait/algebra.proto#L362) using `count-expr`  but Java (Isthmus) is using the older (deprecated, to be fair) and can't cope with the updated version.   

Sounds like a defect in Substrait-Java I think.

nielspardon:
> Sounds like a defect in Substrait-Java I think.

Yes, `FetchRel` was changed in the spec from pure field references to expressions and this change might not have made it to substrait-java yet. Will get to it when the review backlog is cleared on substrait-java.

src/substrait/dataframe/frame.py
```diff
@@ -0,0 +1,949 @@
...
+
+        Every use of the returned frame in the same ``to_plan()`` emits the
+        subplan once as a shared subtree and references it (``ReferenceRel``),
+        instead of inlining a fresh copy each time.
+        """
+        inner = self._plan
+        token = object()  # identity for this cached node
+
+        def resolve(registry: ExtensionRegistry) -> stpl.Plan:
+            ctx = _cte_context.get(None)
```
tokoko: this method feels wrong. I don't really understand the need for contextvars. Can't we simply add a new builder for ReferenceRel in `builders` and use it here, plus modify `infer_rel_schema` to accept subtrees as an optional parameter?

nielspardon: Good call on the `ReferenceRel` construction — I've split that out into a `substrait.builders.plan.reference(...)` builder and `cache()` now uses it, so it's built the same way as every other relation (`fb1c49b`).

On dropping the contextvar via a `subtrees` parameter to `infer_rel_schema`: I tried to talk myself into it but it doesn't actually remove the need. The catch is *where* reference-schema inference gets triggered. It's not just at the top level — it happens deep inside the builder layer. E.g. `base.filter(...)` on a cached frame runs `plan.filter`, which calls `infer_plan_schema(bound_plan)` where `bound_plan`'s root is the `ReferenceRel`. That call site (and the equivalent in `select`/`join`/… — every builder that infers its input schema) has no handle on the subtree list; the only thing that carries it there today is the `reference_subtrees` contextvar.

So a `subtrees=` param would just relocate the problem: each builder would then need the subtree list to pass down, which means widening the `UnboundPlan = Callable[[ExtensionRegistry], Plan]` contract to thread a build-context through every builder. The contextvar is precisely what lets `cache()` add CTE semantics without touching that contract. (There's prior art for this in the repo, too — `outer_schemas` is a contextvar used the same way for correlated-subquery scope.)

Getting to genuinely zero contextvars is doable but it's a builder-layer-wide change (a threaded build context, or a plan-as-AST representation instead of opaque `UnboundPlan` closures). Happy to spike that as a follow-up if you think it's worth it, but I'd rather not grow this PR with it. For now I've documented *why* the contextvar exists at its definition so the next reader isn't left guessing. Does that seem reasonable, or would you prefer I prototype the context-threading approach so we can compare?

tokoko: I agree it's a builder-level change, but probably worth doing rather than introducing contextvar mechanism from data frame that circumvents builder layer.

iiuc plan builder methods always invoke infer_plan_schema with resolved plans objects. substrait Plan objects themselves always contain subtrees (just like extensions). All we have to do is to modify type reference method signatures to be able to pass subtrees down the chain.

I'm fine with leaving all this for a follow-up, but probably best to also simply leave cache/reference out of this PR's scope entirely. We can implement it for both builder and dataframe layers separately.

nielspardon: Good point — you're right, and it's the `reference_subtrees` case specifically that I'd missed: the subtrees ride inside the `Plan` (in `Plan.relations`, like extensions), so the builder-internal `infer_plan_schema(bound_plan)` calls already have them in hand — no contextvar needed once inference takes them as a parameter. That's the clean design.

Agreed on keeping it out of this PR. I've dropped `cache()`/`reference` entirely (the `reference()` builder, the `_cte_context`/`reference_subtrees` contextvars, the `ReferenceRel` inference case, and the tests) so this PR stays focused on the facade, and I've updated the description to match.

Filed #211 to track doing it properly at both layers, with the in-band subtree design written up: type-inference takes `subtrees` as a parameter (from the `Plan`), builders propagate subtrees upward like extensions and rebase `subtree_ordinal`s on multi-input merges (join/set), and `DataFrame.cache()` sits on top of that. Left a note there that the `hint()` guard for `RelCommon`-less relations should come back with it.

src/substrait/dataframe/frame.py
```diff
@@ -0,0 +1,865 @@
+"""The Substrait-native DataFrame.
```
tokoko, line 1: in my mind DataFrame should have been eager unlike builders' closure approach. I'm guessing outer references are the reason for the lazy design, right? Don't have good ideas to handle that, but still noting here.. maybe we should revisit that part of the design eventually. Eager error reporting where possible would probably be a good thing.

tests/dataframe/test_literals.py
```diff
@@ -0,0 +1,256 @@
...
+import pytest
+import substrait.type_pb2 as stt
+
+import substrait.dataframe as sub
+from substrait.builders import type as t
+from substrait.builders.extended_expression import _make_literal
+from substrait.type_inference import infer_literal_type
+
+
+def _built(value, typ):
```
tokoko: this is probably a refactoring artifact

src/substrait/dataframe/frame.py
nielspardon: Yeah, correlated subqueries are the main reason. A subquery captures an inner frame whose `outer_reference`s only resolve once the enclosing query's schema is on the `outer_schemas` stack — which the subquery builder pushes at resolve time, then pops. If the frame resolved eagerly at construction, the inner plan would be built before that outer schema exists, so the outer refs would have nothing to bind to.

It's also partly inherited: the frame is a thin wrapper over the lazy builder closures (`UnboundPlan`), which defer because they take the registry at resolve time. The frame *does* carry a registry from the start, though — so most non-subquery verbs could resolve eagerly and surface errors (unknown column, type mismatch) at the call site instead of at `to_plan()`. A hybrid seems feasible eventually: eager validation for the common path, lazy only for subquery bodies that need an outer context. Agree eager error reporting would be a nice win — filed #212 to track revisiting it.

tests/dataframe/test_literals.py
nielspardon: Yep — `_built` was a leftover one-line wrapper around `_make_literal`; dropped it, the tests call `_make_literal` directly now. Pushed in 81c4723.

tokoko (approved):
looks great, thanks. let's continue with follow-up PRs

PR #220, opened 2026-07-16
refactor: scope extension-relation derivers to the ExtensionRegistry

### LINKED ISSUE #206: Scope extension-relation schema derivers to the ExtensionRegistry instead of a process-global
Follow-up cleanup from review of #204 (non-blocking; not a correctness bug).

`substrait.api`'s user-defined extension relations register their schema deriver in a **process-global** dict rather than on the `ExtensionRegistry` instance:

- `type_inference.register_extension_relation` writes to the module-global `_extension_relation_derivers` (`src/substrait/type_inference.py`), keyed by `type_url`.
- `ExtensionRegistry.register_extension_relation` also writes `self._extension_relations[...]`, but that per-instance dict is **never read** by inference — it's effectively dead state.

Consequences:
- Two `ExtensionRegistry` instances silently share one deriver map; registering a detail class on one changes schema derivation for plans built with another.
- Last-writer-wins if two detail classes share a `type_url`.
- There is no way to unregister or isolate — state leaks across otherwise-independent registries (and across tests).

Every other registry capability is scoped to the instance; extension-relation derivation should be too. Suggested direction: store derivers on the `ExtensionRegistry` and have `_derive_extension_schema` consult the registry already threaded through inference, then drop the global and the dead instance dict.

nielspardon:
## Problem

User-defined extension relations registered their schema deriver in a **process-global** dict in `type_inference` (`_extension_relation_derivers`), while `ExtensionRegistry.register_extension_relation` *also* wrote a per-instance dict (`self._extension_relations`) that inference never read — dead state.

Consequences (from #206):
- Two `ExtensionRegistry` instances silently shared one deriver map.
- Last-writer-wins if two detail classes shared a `type_url`.
- No isolation — state leaked across otherwise-independent registries and across tests.

Every other registry capability is instance-scoped; extension-relation derivation should be too.

## Change

Store derivers only on the `ExtensionRegistry` and thread that registry through inference so `_derive_extension_schema` consults it, then drop the module-global dict and the module-level `register_extension_relation` shim.

- `ExtensionRegistry.register_extension_relation` now writes only `self._extension_relations`; adds `lookup_extension_relation(type_url)`.
- `infer_plan_schema` / `infer_rel_schema` / `infer_expression_type` / `infer_nested_type` / `infer_extended_expression_schema` / `_join_output_struct` take a keyword-only `registry=None` and forward it down to `_derive_extension_schema`. When `registry is None` the behavior is unchanged (leaf/multi raise "no schema deriver"; single passes through), so the many non-extension call sites are unaffected.
- Builder / DataFrame / `expr` call sites pass `registry=registry` — each is already inside a `resolve(registry)` closure, so chained extension-relation inference resolves against the right registry.

## Behavior change

Registration is now scoped to the `ExtensionRegistry` instance. Code following the documented pattern (register on a registry, then build/infer through that same registry — as the `DataFrame` and builders do) is unaffected. Two independent registries no longer share deriver state.

## Testing

- New regression test `test_deriver_registration_is_scoped_to_registry`: a detail class registered on one registry does **not** derive on an independent one (and still derives on the one it was registered on).
- `_kinds` helper now threads the registry (it re-infers a finished plan whose `detail` is an opaque `Any`).
- Full suite: **512 passed, 30 skipped** (unchanged); `ruff format --check` and `ruff check` clean.

Closes #206.

🤖 Generated with AI

### DIFF
--- a/src/substrait/builders/extended_expression.py
+++ b/src/substrait/builders/extended_expression.py
@@ -447,7 +447,8 @@ def scalar_function(
         ]
 
         expression_schemas = [
-            infer_extended_expression_schema(b) for b in bound_expressions
+            infer_extended_expression_schema(b, registry=registry)
+            for b in bound_expressions
         ]
 
         signature = [typ for es in expression_schemas for typ in es.types]
@@ -543,7 +544,8 @@ def aggregate_function(
         ]
 
         expression_schemas = [
-            infer_extended_expression_schema(b) for b in bound_expressions
+            infer_extended_expression_schema(b, registry=registry)
+            for b in bound_expressions
         ]
 
         signature = [typ for es in expression_schemas for typ in es.types]
@@ -640,7 +642,8 @@ def window_function(
         ]
 
         expression_schemas = [
-            infer_extended_expression_schema(b) for b in bound_expressions
+            infer_extended_expression_schema(b, registry=registry)
+            for b in bound_expressions
         ]
 
         signature = [typ for es in expression_schemas for typ in es.types]
--- a/src/substrait/builders/plan.py
+++ b/src/substrait/builders/plan.py
@@ -203,7 +203,7 @@ def project(
 
     def resolve(registry: ExtensionRegistry) -> stp.Plan:
         _plan = plan if isinstance(plan, stp.Plan) else plan(registry)
-        ns = infer_plan_schema(_plan)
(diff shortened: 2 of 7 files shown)


tokoko (approved):
lgtm

Issue #186, opened 2026-06-17
add `ExecutionContextVariable` support (`CURRENT_DATE`, `CURRENT_TIMESTAMP`, `CURRENT_TIMEZONE`)


nielspardon:

nielspardon:
Implemented by #204 (spec bump to 0.96.0). `ExecutionContextVariable` (`CURRENT_DATE`/`CURRENT_TIMESTAMP`/`CURRENT_TIMEZONE`) is fully wired:

- builder: `builders/extended_expression.py` `execution_context_variable()`;
- DataFrame API: `dataframe/expr.py` `current_timestamp` / `current_date` / `current_timezone` (exported in `dataframe/__init__.py`);
- type inference: `type_inference.py` returns `precision_timestamp_tz` / `date` / `string` per variant.

Covered by parametrized tests in `tests/dataframe/test_frame.py` (all 3 pass).

PR #224, opened 2026-07-17
feat: add Plan.ExecutionBehavior support

### LINKED ISSUE #185: add new `Plan.ExecutionBehavior`

nielspardon:
## What

Adds support for the plan-level **execution behavior** introduced in Substrait 0.87.0 (`Plan.execution_behavior.variable_eval_mode`). This is the last open sub-issue of #161; the companion `ExecutionContextVariable` support (#186) is already merged.

`variable_eval_mode` controls how often execution context variables (`current_timestamp` / `current_date` / `current_timezone`) are evaluated — once per plan (`VARIABLE_EVALUATION_MODE_PER_PLAN`) or once per record (`VARIABLE_EVALUATION_MODE_PER_RECORD`).

## Changes

- **`builders.plan.with_execution_behavior(plan, variable_eval_mode)`** — sets `Plan.execution_behavior` on a copy of the plan.
- **Propagation** — execution behavior is now carried over by every relational builder. Rather than sprinkle a call into ~19 builders, it is folded into the single metadata choke point every builder already routes its inputs through (`_merge_extensions`, renamed to `_merge_plan_metadata`). As a result the setting survives subsequent operations and is **order-independent** across a pipeline, instead of only surviving as the last step.
- **`DataFrame.with_execution_behavior("per_plan" | "per_record")`** — wires it into the native fluent frame, following the existing string→enum convention (`_JOIN_TYPES`, `_CREATE_MODES`, …).

## Test plan

- New `tests/builders/plan/test_with_execution_behavior.py`: set/preserve-relations/no-input-mutation, propagation through a later builder, first-input-wins for multi-input builders, last-wins overwrite, unset-by-default.
- New DataFrame cases in `tests/dataframe/test_frame.py`: both modes, preserved-through-later-ops, set mid-chain, invalid-mode error, and an end-to-end pairing with `current_timestamp`.
- Full suite: 536 passed, 30 skipped; `ruff check` and `ruff format --check` clean.

Resolves #185.

🤖 Generated with AI

### DIFF
--- a/src/substrait/dataframe/frame.py
+++ b/src/substrait/dataframe/frame.py
@@ -523,20 +523,20 @@ class DataFrame:
 
         return self._next(resolve)
 
-    def with_execution_behavior(self, mode: str) -> "DataFrame":
+    def with_execution_behavior(self, variable_eval_mode: str) -> "DataFrame":
         """Set how often execution context variables are evaluated (plan-level).
 
-        ``mode`` is ``"per_plan"`` (evaluate once for the whole plan) or
-        ``"per_record"`` (evaluate once per record), controlling variables such
-        as :func:`substrait.dataframe.current_timestamp`. The setting is carried
-        across subsequent operations, so it may be applied at any point in the
-        chain.
+        ``variable_eval_mode`` is ``"per_plan"`` (evaluate once for the whole
+        plan) or ``"per_record"`` (evaluate once per record), controlling
+        variables such as :func:`substrait.dataframe.current_timestamp`. The
+        setting is carried across subsequent operations, so it may be applied at
+        any point in the chain.
         """
         try:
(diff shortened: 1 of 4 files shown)


src/substrait/dataframe/frame.py
```diff
@@ -516,6 +523,24 @@ def resolve(registry: ExtensionRegistry):
 
         return self._next(resolve)
 
+    def with_execution_behavior(self, mode: str) -> "DataFrame":
```
tokoko: `mode` feels too generic here, maybe `variable_eval_mode`?

nielspardon: Good call — renamed to `variable_eval_mode` in ae055ca, which also lines it up with the `builders.plan.with_execution_behavior` parameter name.

tokoko (approved):

PR #259, opened 2026-09-01
feat: implement enum function arguments

### LINKED ISSUE #258: Enumeration-argument functions (`extract`, `round_temporal`, …) cannot be built
Standard functions that take **enumeration arguments** (`extract`,
`round_temporal`, etc.) cannot be authored through the DataFrame /
builder API. Resolution fails, and even when forced past resolution the
enum selection is serialized into the wrong field, producing a plan that
downstream consumers (e.g. DuckDB) reject.

# Background

The Substrait spec declares enumeration arguments under `args:` with an
`options:` domain rather than a `value:` type. `extract` is the canonical
example (`functions_datetime.yaml`):

```yaml
- name: "extract"
  impls:
    - args:
        - name: component        # enumeration argument
          options: [ YEAR, ISO_YEAR, US_YEAR, UNIX_TIME ]
        - name: x                # value argument
          value: date
      return: i64
```

Per the spec, enumeration selections serialize into
`ScalarFunction.arguments` as `FunctionArgument.enum`, **interleaved with the
value arguments in signature order**. They do *not* belong in
`ScalarFunction.options`, which is reserved for *behavioral* options such as
`overflow` and `rounding`.

## Reproduction

```python
import substrait.dataframe as sub

df = sub.read_named_table("t", {"d": sub.date})
plan = df.with_columns(y=sub.f.extract(sub.col("d"), component="YEAR")).to_plan()
```

## Expected behavior

The plan builds, and the resulting `ScalarFunction` carries:

- `arguments = [FunctionArgument(enum="YEAR"), FunctionArgument(value=<col d>)]`
  — the enum selection in `arguments`, in declared signature order;
- `options = []` — nothing leaked into behavioral options.

## Actual behavior

Two distinct defects:

1. **Resolution fails.** The overload's arity counts the enum positions
   (`extract` is arity 2: `[component, x]`), but the signature handed to the
   registry contains only the value operands (`[date]`). The arity check never
   matches, so resolution raises:

   ```
   Exception: Unknown function extract for [<date type>]
   ```

2. **Serialization is wrong.** Even once resolution is bypassed, the enum
   selection arrives via the `**options` keyword path and is emitted as a
   `FunctionOption`. This omits the enum from `arguments` entirely. A consumer
   such as DuckDB reads enum selections exclusively from `arguments`, so it sees
   an empty enum vector and crashes (`index 0 within vector of size 0`).

## Root cause

- The extension registry matches an overload against a value-only signature and
  has no way to account for enumeration positions when the selections are
  provided by name rather than interleaved into the signature.
- The DataFrame function helpers collect *all* keyword arguments into the
  behavioral `options` dict, with no distinction between an enumeration-argument
  selection (`component="YEAR"`) and a genuine behavioral option
  (`overflow="ERROR"`).
- The builders (`scalar_function` / `aggregate_function` / `window_function`)
  emit `arguments` from the value operands only and pass the entire `options`
  dict to `FunctionOption`.

## Affected versions

Reproduces on 0.31.0 (and earlier). The registry has latent enum awareness
(`satisfies_signature` already handles an enum string interleaved into the
signature), but no caller in the DataFrame/builder pipeline exercises it, so the
end-to-end path is broken.

## Environment

- substrait-python 0.31.0
- Python 3.12 / 3.13

thgschwind:
Enumeration-argument functions such as extract and round_temporal declare arguments (component, indexing) with an options: domain rather than a value type, and Substrait counts them toward the overload's arity. They could not be authored through the DataFrame / builder API.

Resolution was handed a value-only signature, so the arity never matched the overload and the build raised Unknown function extract. Forced past that, the selection was emitted as a FunctionOption. Since the spec serializes enumeration selections into arguments as FunctionArgument.enum, interleaved with the value arguments in declared order, and reserves options for behavioral settings such as overflow and rounding, a consumer that reads enums only from arguments, such as DuckDB, then sees an empty enum vector and crashes.

Enumeration selections are now supplied positionally with a sub.enum(...) marker, written in the argument slot the spec assigns them:

```
sub.f.extract(sub.enum("YEAR"), sub.col("d"))
sub.f.extract(sub.enum("DAY"), sub.enum("ONE"), sub.col("ts"))
sub.f.median(sub.enum("EXACT"), sub.col("x"), rounding="TRUNCATE")
```

This mirrors substrait-go's types.Enum and substrait-java's EnumArg, which both pass the selection positionally in the argument list and keep behavioral options on a separate channel. The builders walk the argument list once, emitting FunctionArgument.enum for each marker and FunctionArgument.value for each expression, and the registry matches the enum token positionally against the overload's option domain — so the selection counts toward arity as the spec requires. **options now carries only behavioral options and does not participate in overload matching.

Fixes #258

### DIFF
--- a/src/substrait/builders/extended_expression.py
+++ b/src/substrait/builders/extended_expression.py
@@ -83,6 +83,36 @@ def _function_options(options):
     return result
 
 
+def _function_arguments(func_entry, bound_expressions, options):
+    """Interleave value operands and enumeration selections into FunctionArguments.
+
+    Value operands come from ``bound_expressions`` (positional), enumeration
+    selections from ``options`` (by argument name). ``func_entry`` weaves them
+    into declared signature order -- enum selections as ``FunctionArgument.enum``,
+    values as ``FunctionArgument.value``. Returns ``(arguments, options)`` where
+    the second element is the options *not* consumed as enumeration arguments
+    (i.e. the behavioral options destined for ``FunctionOption``).
+    """
+    value_args = (
+        stalg.FunctionArgument(value=e.referred_expr[0].expression)
+        for e in bound_expressions
+    )
+    plan = func_entry.interleave_arguments(value_args, options)
+    if plan is None:
+        # Resolution already matched this overload, so interleaving should not
+        # fail; fall back to the value-only form rather than lose the operands.
+        return [
+            stalg.FunctionArgument(value=e.referred_expr[0].expression)
+            for e in bound_expressions
+        ], options
+    ordered, behavioral_options = plan
+    arguments = [
+        stalg.FunctionArgument(enum=selection) if kind == "enum" else selection
+        for kind, selection in ordered
+    ]
+    return arguments, behavioral_options
+
+
 def resolve_expression(
     expression: ExtendedExpressionOrUnbound,
     base_schema: stp.NamedStruct,
@@ -581,12 +611,16 @@ def scalar_function(
 
         signature = [typ for es in expression_schemas for typ in es.types]
 
-        func = registry.lookup_function(urn, function, signature)
+        func = registry.lookup_function(urn, function, signature, options=options)
 
         if not func:
             raise Exception(f"Unknown function {function} for {signature}")
 
         func_ref = function_reference(urn, str(func[0]))
+
+        arguments, behavioral_options = _function_arguments(
+            func[0], bound_expressions, options
+        )
 
         return stee.ExtendedExpression(
             referred_expr=[
@@ -594,13 +628,8 @@ def scalar_function(
                     expression=stalg.Expression(
                         scalar_function=stalg.Expression.ScalarFunction(
                             function_reference=func_ref,
-                            arguments=[
-                                stalg.FunctionArgument(
-                                    value=e.referred_expr[0].expression
-                                )
-                                for e in bound_expressions
-                            ],
-                            options=_function_options(options),
+                            arguments=arguments,
+                            options=_function_options(behavioral_options),
                             output_type=func[1],
                         )
                     ),
@@ -654,23 +683,24 @@ def aggregate_function(
 
         signature = [typ for es in expression_schemas for typ in es.types]
 
-        func = registry.lookup_function(urn, function, signature)
+        func = registry.lookup_function(urn, function, signature, options=options)
 
         if not func:
             raise Exception(f"Unknown function {function} for {signature}")
 
         func_ref = function_reference(urn, str(func[0]))
 
+        arguments, behavioral_options = _function_arguments(
+            func[0], bound_expressions, options
+        )
+
         return stee.ExtendedExpression(
             referred_expr=[
                 stee.ExpressionReference(
                     measure=stalg.AggregateFunction(
                         function_reference=func_ref,
-                        arguments=[
-                            stalg.FunctionArgument(value=e.referred_expr[0].expression)
-                            for e in bound_expressions
-                        ],
-                        options=_function_options(options),
+                        arguments=arguments,
+                        options=_function_options(behavioral_options),
                         output_type=func[1],
                         invocation=invocation
                         if invocation is not None
@@ -724,12 +754,16 @@ def window_function(
 
         signature = [typ for es in expression_schemas for typ in es.types]
 
-        func = registry.lookup_function(urn, function, signature)
+        func = registry.lookup_function(urn, function, signature, options=options)
 
         if not func:
             raise Exception(f"Unknown function {function} for {signature}")
 
         func_ref = function_reference(urn, str(func[0]))
+
+        arguments, behavioral_options = _function_arguments(
+            func[0], bound_expressions, options
+        )
 
         return stee.ExtendedExpression(
             referred_expr=[
@@ -737,13 +771,8 @@ def window_function(
                     expression=stalg.Expression(
                         window_function=stalg.Expression.WindowFunction(
                             function_reference=func_ref,
-                            arguments=[
-                                stalg.FunctionArgument(
-                                    value=e.referred_expr[0].expression
-                                )
-                                for e in bound_expressions
-                            ],
-                            options=_function_options(options),
+                            arguments=arguments,
+                            options=_function_options(behavioral_options),
                             output_type=func[1],
                             partitions=[
                                 e.referred_expr[0].expression for e in bound_partitions
--- a/src/substrait/dataframe/expr.py
+++ b/src/substrait/dataframe/expr.py
@@ -251,7 +251,7 @@ def _resolve_over_urns(
         for b in bound
         for typ in infer_extended_expression_schema(b, registry=registry).types
     ]
-    match = registry.find_function(name, signature, urns)
+    match = registry.find_function(name, signature, urns, options=options)
     if match is not None:
         winning_urn = match[0].urn
         return builder(
--- a/src/substrait/extension_registry/function_entry.py
+++ b/src/substrait/extension_registry/function_entry.py
@@ -10,6 +10,8 @@ from substrait.derivation_expression import _parse, evaluate
 
 from .signature_checker_helpers import covers, normalize_substrait_type_names
 
+_MISSING = object()
+
 
 class FunctionType(Enum):
     SCALAR = "scalar"
@@ -31,6 +33,10 @@ class FunctionEntry:
         self.urn: str = urn
         self.function_type = function_type
         self.arguments = []
+        # Argument names parallel to ``arguments`` / ``normalized_inputs``. Used to
+        # match enumeration selections, which arrive by name (e.g. component="YEAR")
+        # rather than by position, back to their declared signature slot.
+        self.arg_names: list = []
         self.nullability = (
             impl.nullability if impl.nullability else se.NullabilityHandling.MIRROR
         )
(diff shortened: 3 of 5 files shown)


src/substrait/extension_registry/function_entry.py
```diff
@@ -41,24 +47,97 @@ def __init__(
...
+            if kind == "req" and name in remaining:  # enum selection by name
+                interleaved.append(str(remaining.pop(name)))
+                continue
+            item = next(items, _MISSING)
+            if item is _MISSING:
+                return None
+            interleaved.append(item)
+        if next(items, _MISSING) is not _MISSING:
+            return None  # more operands than the signature declares
+        return interleaved
```
nielspardon: Return `remaining` alongside `interleaved` and have `satisfies_signature` reject the overload when a leftover name is an enumeration argument of the function. Discarding it lets `std_dev:fp64` — deprecated in spec 0.88.0 with "Use implementation with distribution enum argument instead" — win over `std_dev:req_fp64` on YAML declaration order, which is why #161's "use enum arg function signatures for `std_dev` and `variance`" is still open and why a stray `indexing="ONE"` becomes a `FunctionOption`. Ranking matches by how many enum selections they consume would work too — worth deciding deliberately, because as written an upstream YAML reordering changes the emitted plan for unchanged user code.

src/substrait/dataframe/expr.py
```diff
@@ -251,7 +251,7 @@ def _resolve_over_urns(
         for b in bound
         for typ in infer_extended_expression_schema(b, registry=registry).types
     ]
-    match = registry.find_function(name, signature, urns)
+    match = registry.find_function(name, signature, urns, options=options)
```
nielspardon: Please say in the PR body why enum selections share the `**options` keyword channel rather than being positional — spec v0.99.0 separates the two deliberately (`scalar_functions.md:73-80`: enum args "Positional, part of the argument list", options "Named, separate from arguments"), and both sibling bindings with a builder keep them apart: substrait-java's `EnumArg` is a first-class positional `FunctionArg`, and substrait-go's `types.Enum` is a positional `FuncArg` with options as a separate parameter. Sharing the channel is what makes `distribution` ambiguous on `std_dev`/`variance`, and name-keying is fragile besides — an enum arg's `name` is optional, defaulting to a positional name like `arg0`. The API is the expensive thing to change after merge, so it's worth recording the choice even if you keep it.

src/substrait/extension_registry/function_entry.py
```diff
@@ -41,24 +47,97 @@ def __init__(
...
+                return None
             inputs = self.arguments
-        if len(inputs) != len(signature):
+        if len(inputs) != len(interleaved):
             return None
-        zipped_args = list(zip(inputs, signature))
+        zipped_args = list(zip(inputs, interleaved))
         parameters = {}
         for x, y in zipped_args:
             if isinstance(y, str):
```
nielspardon, line 75: Reject an enumeration slot that received a value operand rather than a selection: an enum slot's `x` is the options list, so `covers()` calls `.typeDef()` on a `list` and the `AttributeError` escapes `lookup_function`. `sub.f.round_temporal(col("ts"), col("i"), col("ts"), col("ts"), rounding="FLOOR")` raises it here where `main` returned a clean `Unknown function round_temporal`.
```suggestion
        for x, y in zipped_args:
            if isinstance(x, list) and not isinstance(y, str):
                return None
            if isinstance(y, str):
```

nielspardon (commented):
Make `_resolve_signature` return its unconsumed `options` and reject a match that leaves an enumeration-argument name unclaimed — `sub.f.std_dev(col, distribution="SAMPLE")` and `variance` still resolve to the value-only overload and emit `distribution` as a `FunctionOption`, byte-identical to the pre-PR plan, so #161's unticked enum-arg item for those two functions survives this PR. The same discarded leftover lets a stray selection through: `extract(col, component="YEAR", indexing="ONE")` builds `extract:req_date` and puts `indexing` in `options`. Four inline notes: that fix, the keyword-channel choice behind it, a resolution crash, and an emission fallback.

src/substrait/builders/extended_expression.py
```diff
@@ -83,6 +83,36 @@ def _function_options(options):
...
+    values as ``FunctionArgument.value``. Returns ``(arguments, options)`` where
+    the second element is the options *not* consumed as enumeration arguments
+    (i.e. the behavioral options destined for ``FunctionOption``).
+    """
+    value_args = (
+        stalg.FunctionArgument(value=e.referred_expr[0].expression)
+        for e in bound_expressions
+    )
+    plan = func_entry.interleave_arguments(value_args, options)
+    if plan is None:
```
nielspardon: Raise here instead of falling back — the fallback returns `options` unfiltered, so the selection is emitted as a `FunctionOption` and dropped from `arguments`, which is the shape this PR exists to remove. It is reachable rather than defensive: `scalar_function(datetime_urn, "extract", expressions=[ee], options={"component": "YEAR"})` with an `ee` carrying two `referred_expr` resolves `extract:req_ptstz_str`, then emits `arguments=['value']` with `component` in `options` and no error. `signature` counts types while the operand list counts bound expressions, so the two arities can diverge.

nielspardon (commented):
Round 2 — the tail of the review. One regression (`quantile`), three smaller gaps, the test coverage that would have caught them, and one grouped note for two spots outside this diff that the new `FunctionArgument.enum` output breaks.

src/substrait/extension_registry/function_entry.py
```diff
@@ -41,24 +47,97 @@ def __init__(
...
                 return None
             inputs = [self.arguments[0]] * len(signature)
+            interleaved: list = list(signature)
         else:
+            interleaved = self._resolve_signature(signature, options)
+            if interleaved is None:
+                return None
             inputs = self.arguments
-        if len(inputs) != len(signature):
+        if len(inputs) != len(interleaved):
```
nielspardon: Reject an overload whose return type cannot be evaluated instead of letting the exception escape. `sub.f.quantile(col("i"), col("x"), boundaries="BOTH", precision="EXACT")` raises a bare `Exception('')` from `derivation_expression` where `main` returned a clean `Unknown function quantile for [...]` — filling the enum slots from `options` makes the arity match for the first time, so `evaluate()` reaches `quantile`'s `LIST?<any>` return. `_find_matching_functions` has no `try`/`except`, so one unevaluable overload aborts the whole lookup even when a later one would match.

src/substrait/extension_registry/function_entry.py
```diff
@@ -41,24 +47,97 @@ def __init__(
...
+        prefers ``options`` and otherwise consumes the next ``signature`` item.
+        Returns the value-and-enum sequence to match against ``self.arguments``,
+        or ``None`` on an arity mismatch.
+        """
+        remaining = dict(options or {})
+        items = iter(signature)
+        interleaved: list = []
+        for kind, name in zip(self.normalized_inputs, self.arg_names):
+            if kind == "req" and name in remaining:  # enum selection by name
+                interleaved.append(str(remaining.pop(name)))
```
nielspardon: Validate that a selection is a single string, and name it in the error. `str(remaining.pop(name))` turns `component=["YEAR"]` into `"['YEAR']"`, which matches no domain, so the user gets `Unknown function extract for [date ...]` pointing at the operand types rather than at `component` — and `_function_options` documents a list as a legal preference for this very same kwarg. Casing is worth a decision here too: spec v0.99.0 makes `enum: "year"` a *valid plan* (`algebra.proto:1314`), so rejecting `component="year"` is defensible, but the message shouldn't imply it isn't an allowed option.

src/substrait/extension_registry/function_entry.py
```diff
@@ -41,24 +47,97 @@ def __init__(
                     self.normalized_inputs.append(
                         normalize_substrait_type_names(arg.value)
                     )
+                    self.arg_names.append(arg.name)
                 elif isinstance(arg, se.EnumerationArg):
                     self.arguments.append(arg.options)
                     self.normalized_inputs.append("req")
+                    self.arg_names.append(arg.name)
```
nielspardon: Fall back to the positional default when an enumeration argument has no name — spec v0.99.0 says `name` is "Optional, defaults to a name based on position (e.g. `arg0`)" (`scalar_functions.md:49`), but `arg.name` can be `None` here and `None not in remaining` then makes the overload unreachable by keyword with no diagnostic. No standard extension hits this today; a custom YAML registered through `register_extension_yaml` can.

src/substrait/extension_registry/registry.py
```diff
@@ -141,9 +148,12 @@ def lookup_function(
...
+        options: Optional[dict] = None,
     ) -> Optional[tuple[FunctionEntry, Type]]:
         """Look up a function within a specific URN."""
-        matches = self._find_matching_functions(function_name, signature, [urn])
+        matches = self._find_matching_functions(
+            function_name, signature, [urn], options
+        )
         return matches[0] if matches else None
 
     def list_functions(
```
nielspardon, line 153: Thread `options` through `list_functions` and `list_functions_across_urns` (line 165) as well — they call `_find_matching_functions` with `options=None`, so they return `[]` for every enumeration-argument overload while `lookup_function` and `find_function` match it. `find_function`'s docstring still claims it generalizes `list_functions_across_urns`, which is no longer true.

tests/dataframe/test_enum_arguments.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Must not raise "Unknown function extract".
+    plan = df.with_columns(y=sub.f.extract(sub.col("d"), component="YEAR")).to_plan()
+
+    (extract,) = [
+        sf
+        for sf in _scalar_functions(plan)
+        if _function_name(plan, sf).startswith("extract")
+    ]
+    assert extract, "extract did not resolve to a registry function"
```
nielspardon: Drop this assertion and add the cases that would have caught the gaps: protobuf messages define no `__bool__`, so `assert extract` is always true, and the unpack above already fails on zero or multiple matches. `std_dev`/`variance`, the aggregate and window call sites this PR changed, and a call carrying both kinds at once (`median(col, precision="EXACT", rounding="TRUNCATE")` — the split that is the PR's whole premise) are all untested. Note `_function_name(...).startswith("extract")` also matches `extract_boolean`, so extending the suite in place will trip the single-match unpack.
```suggestion
```

src/substrait/builders/extended_expression.py
```diff
@@ -83,6 +83,36 @@ def _function_options(options):
...
+        # Resolution already matched this overload, so interleaving should not
+        # fail; fall back to the value-only form rather than lose the operands.
+        return [
+            stalg.FunctionArgument(value=e.referred_expr[0].expression)
+            for e in bound_expressions
+        ], options
+    ordered, behavioral_options = plan
+    arguments = [
+        stalg.FunctionArgument(enum=selection) if kind == "enum" else selection
```
nielspardon: Two things outside this diff that the new `FunctionArgument.enum` output breaks — no suggestions attachable, so pointers only. `src/substrait/utils/display.py:657` and `:707` gate on `arg.HasField("value")` and then test `arg.value.HasField("enum")`, but `enum` is a `FunctionArgument` field and `Expression` has none — so `pretty_print_plan` renders every selection as `<function_argument>`, and that check is both dead and a `ValueError` if ever reached; `arg.HasField("enum")` is the fix. `src/substrait/builders/extended_expression.py:636` feeds `_alias_or_inferred` only the value operands, so `select(extract(d, component="YEAR"), extract(d, component="ISO_YEAR"))` yields `['extract(d)', 'extract(d)']` — two different projections colliding on one name, unreachable before this PR.

src/substrait/extension_registry/function_entry.py
thgschwind: You're right that the leftover-options mechanism was the root problem, and rather than rank matches by consumed selections we removed the mechanism entirely: enum selections are now positional sub.enum(...) markers in the signature in declared order.  Resolution is pure arity/position again. std_dev/variance now select the enum overload because sub.f.std_dev(sub.enum("POPULATION"), col) presents two argument slots and the deprecated value-only overload presents one — no YAML-order sensitivity, and #161's item for those two is genuinely satisfiable. A stray indexing can no longer become a FunctionOption because there is no name-keyed options channel feeding resolution.

src/substrait/dataframe/expr.py
thgschwind: Agreed.  We've reworked enum arguments to be positional, matching the substrait-java's EnumArg and substrait-go's types.Enum implementation as well as the spec's "positional, part of the argument list".  **options now carries only behavioral options. I will update the PR's body accordingly.

src/substrait/extension_registry/function_entry.py
thgschwind: Fixed. We guard the slot mismatch in both directions now — a value Type in an enum slot and an enum token in a value slot both return None cleanly instead of letting covers() call .typeDef() on the options list: if isinstance(x, list) != isinstance(y, str): return None. round_temporal(...) with the wrong arity again returns a clean Unknown function.

src/substrait/builders/extended_expression.py
thgschwind: The fallback is gone entirely. In the positional design _function_arguments walks the call's arguments once, emitting FunctionArgument.enum for a marker and FunctionArgument.value for an operand — there's no re-interleaving from options and therefore no path that can silently drop a selection into FunctionOption. The arity divergence you spotted (types vs. bound expressions) no longer exists because the same positional list drives both signature and arguments.

src/substrait/extension_registry/function_entry.py
thgschwind: Fixed independently of the channel change, since it can bite any overload whose arity happens to match: evaluate(self.impl.return_, parameters) is now wrapped in try/except, returning None so a later overload can still match instead of the exception aborting the whole _find_matching_functions loop. quantile(...) returns a clean Unknown function again.

src/substrait/extension_registry/function_entry.py
thgschwind: With the positional marker this is enforced at construction: sub.enum(value) raises ValueError if value isn't a str, so sub.enum(["YEAR"]) fails immediately and names enum() rather than surfacing later as a misleading Unknown function. On casing: we match the token against the overload's domain as-is, so a wrong-case token fails resolution like any other signature mismatch; we deliberately don't normalize case yet (conservative — enum: "year" being a valid plan means we can relax this later without a breaking change).

src/substrait/extension_registry/function_entry.py
thgschwind: No longer applicable — resolution is by position, not by name, so an enum argument with no name in a custom YAML resolves fine. We dropped the arg_names tracking that made the name load-bearing.

src/substrait/extension_registry/registry.py
thgschwind: Resolved by removing options from the matching path altogether. All four entry points (lookup_function, find_function, list_functions, list_functions_across_urns) now share _find_matching_functions on a pure positional signature, so the asymmetry is gone and find_function's "generalizes list_functions_across_urns" docstring is accurate again.

tests/dataframe/test_enum_arguments.py
thgschwind: Done. The assert extract (which, as you note, protobuf messages make always-true) is gone; single-match is enforced by tuple-unpacking a helper that matches the exact function name, so extract no longer collides with extract_boolean. Added cases for the previously untested paths: std_dev selecting the enum overload by arity, the aggregate and window call sites, and median(sub.enum("EXACT"), col, rounding="TRUNCATE") — the enum-arg/behavioral-option split that is the PR's whole premise — plus two distinct extract selections asserting distinct inferred output names.

src/substrait/builders/extended_expression.py
thgschwind: Both fixed. display.py:657/707 now gate on arg.HasField("enum") at the FunctionArgument level (the old arg.value.HasField("enum") was dead and would have raised), so selections render instead of <function_argument>. And _alias_or_inferred is now fed the enum tokens alongside the value operands, so select(extract(sub.enum("YEAR"), d), extract(sub.enum("ISO_YEAR"), d)) yields two distinct inferred names rather than colliding on one.

andrew-coleman (approved):

PR #266, opened 2026-09-07
fix!: apply ReadRel projection during schema inference

### LINKED ISSUE #264: infer_rel_schema ignores ReadRel.projection
DuckDB 1.5.5 (substrait extension 4ea63dd) and DataFusion 54.0.0 put a projection mask on every table read in the plans I got from them (42 of 42 each), so this reaches their plans whenever the mask drops or reorders columns. For DuckDB's plan of `SELECT c2, c0 FROM t`, whose read carries the mask `[2, 0]`, `infer_rel_schema` returns three types for two root names.

`infer_rel_schema`'s read branch takes `rel.read.base_schema.struct` and never reads `rel.read.projection` (`type_inference.py:698`). The word `projection` does not appear anywhere in that file, on `main` either, so a masked read infers the columns the mask left out.

[Read Operator](https://substrait.io/relations/logical_relations/#read-operator), Direct Output Order: *"Defaults to the schema of the data read after the optional projection (masked complex expression) is applied."*

Reproduced with 0.31.0:

```python
import substrait.algebra_pb2 as stalg
import substrait.type_pb2 as stt
from substrait.type_inference import infer_rel_schema

i64 = stt.Type(i64=stt.Type.I64(nullability=stt.Type.NULLABILITY_REQUIRED))
rel = stalg.Rel(read=stalg.ReadRel(
    base_schema=stt.NamedStruct(names=["a", "b", "c"],
        struct=stt.Type.Struct(types=[i64, i64, i64],
                               nullability=stt.Type.NULLABILITY_REQUIRED)),
    projection=stalg.Expression.MaskExpression(
        select=stalg.Expression.MaskExpression.StructSelect(struct_items=[
            stalg.Expression.MaskExpression.StructItem(field=2)]))))

print(len(infer_rel_schema(rel).types))
# actual:   3
# expected: 1, the one column the mask selects
```

Foreign plans only, as far as I can see: nothing in `builders/` or `dataframe/` sets `ReadRel.projection`, so this is reachable the same way #217 was rather than through the library's own API.

The narrowing would apply to `base_schema.names` as well as to `struct.types`, and it belongs before `common`/emit, whose indices are defined over the post-projection order. One scope question worth settling first: the repro uses a flat `StructSelect`, and masks also nest through `struct_items.child`, so it is worth deciding whether the fix covers the nested form now or later. Happy to send the PR.

alexandrefimov:
`infer_rel_schema` currently returns the full `base_schema` for a masked read. A mask selecting one field can therefore report all input fields, and emit mappings are evaluated against the wrong columns.

Apply the read projection before emit, following the [Read Operator in spec v0.99.0](https://github.com/substrait-io/substrait/blob/v0.99.0/site/docs/relations/logical_relations.md#read-operator). Recursively project struct fields, list elements, and map values while preserving their nullability and container types. Keep explicit `RelRoot.names` and the original input schema unchanged. For projected reads, leave correlations inside `filter` and `best_effort_filter` offset-based: [Read Filtering](https://github.com/substrait-io/substrait/blob/v0.99.0/site/docs/relations/logical_relations.md#read-filtering) explicitly places those predicates before projection.

Mask unwrapping is a compatibility choice. The [unwrapping section](https://github.com/substrait-io/substrait/blob/v0.99.0/site/docs/expressions/field_references.md#unwrapping-behavior) describes singleton unwrapping by default but leaves its serialization as TBD. This change follows Java's type projector: it preserves structs and containers at every level regardless of `maintain_singular_struct`. For example, selecting only the string from a nested `struct<i64, string>` keeps a nested `struct<string>`, even with the flag false. The validator instead requires the flag for a single-field read projection.

Struct fields also follow Java's mask order, including reordered and repeated selections. The same spec page still lists column reordering as an open question, so preserving mask order and duplicates remains a compatibility choice.

Closes #264

BREAKING CHANGE: ReadRel projection masks now change inferred output schemas and the indices seen by emit and parent relations. Invalid mask field indices and selector/type mismatches now raise errors instead of being ignored.

## Summary by CodeRabbit

- **Bug Fixes**
  - Read projections now preserve field ordering and nested structure metadata.
  - Nested structs, lists, and maps are projected correctly while retaining their surrounding types.
  - Invalid field indexes and incompatible nested selections now produce clear errors.
  - Read filters continue to reference the correct underlying schema when projections are applied.
  - Downstream expressions now use projected field indexes consistently.

- **Tests**
  - Added comprehensive coverage for nested projections, filters, invalid selections, and plan immutability.


src/substrait/type_inference.py
```diff
@@ -696,6 +746,12 @@ def infer_rel_schema(rel: stalg.Rel, *, registry=None, subtrees=()) -> stt.Type.
 
     if rel_type == "read":
         (common, struct) = (rel.read.common, rel.read.base_schema.struct)
+        if rel.read.HasField("projection"):
+            # Spec v0.99.0 describes default unwrapping but leaves its
+            # serialization open. Match Java by keeping structs and containers
+            # at every level, regardless of maintain_singular_struct.
```
tokoko: this sounds like a bug in java impl, right? why should we ignore `maintain_singular_struct`?

alexandrefimov: Right, `MaskExpressionTypeProjector` never consults it. The flag switches unwrapping on top of the field selection this PR applies: with `true` we already do what it asks, and `false` would only matter below the top level, since `infer_rel_schema` has to return the read's row as a `Type.Struct`. Reworded that comment in 95a2e9a to rest on the schema rather than on Java, and to say that nested levels keep the struct too.

src/substrait/utils/__init__.py
```diff
@@ -604,15 +607,18 @@ def convert_rel(rel, scope):
         if node is not None:
             # The relation whose output row a subquery here would see one level up:
             # a single-input host exposes its input; a leaf or multi-input host its
-            # own output -- except a reducing join's combined-inputs-scoped fields,
-            # whose scope no relation's output carries (binding None -> left as-is).
+            # own output. Join conditions and projected read filters may use a
+            # different row with no relation to anchor (binding None -> left as-is).
             single_input = _child_rel(*children[0]) if len(children) == 1 else None
             reducing = single_input is None and _is_reducing_join(node)
+            projected_read = rel_type == "read" and node.HasField("projection")
```
nielspardon: Include `common.emit` here. A read's `filter` binds against its base schema, and an emit hides or permutes the output row just as a projection does, so an emit-only read still gets anchored to a row it does not carry — a 3-column read with `emit=[0]` and a correlated subquery in `filter` referencing field 2 gets rewritten to `rel_reference` and then resolves against a 1-field struct.

```suggestion
            projected_read = rel_type == "read" and (
                node.HasField("projection")
                or node.common.WhichOneof("emit_kind") == "emit"
            )
```

nielspardon (commented):
Widen the new `projected_read` guard to cover `common.emit` as well — an emit reshapes a read's output row exactly as a projection does, so the mis-anchoring this PR fixes for masks is still live for emits, and the new test only parametrizes projected/unprojected.

Separately, move the `BREAKING CHANGE:` footer back to the end of the description. The CodeRabbit block now trailing it is absorbed into the published note, so the whole summary (comment markers included) lands in the ⚠ BREAKING CHANGES entry in `CHANGELOG.md`.

src/substrait/type_inference.py
```diff
@@ -696,6 +746,13 @@ def infer_rel_schema(rel: stalg.Rel, *, registry=None, subtrees=()) -> stt.Type.
 
     if rel_type == "read":
         (common, struct) = (rel.read.common, rel.read.base_schema.struct)
+        if rel.read.HasField("projection"):
```
nielspardon: Gate this on `projection.HasField("select")`. A `MaskExpression` can be present with `select` unset — assigning `maintain_singular_struct` alone is enough to mark the parent present — and the mask then selects nothing, so the read collapses to zero fields where it previously returned the full base schema. The spec makes an absent projection default to all of schema, never to none.

```suggestion
        if rel.read.projection.HasField("select"):
```

src/substrait/utils/__init__.py
alexandrefimov: Done in f3fe470. The read-filter test now also runs on an emit-only read.

src/substrait/type_inference.py
alexandrefimov: Done in f3fe470, with a test for a mask that sets only `maintain_singular_struct`.

src/substrait/utils/__init__.py
```diff
@@ -604,15 +607,22 @@ def convert_rel(rel, scope):
...
-            # whose scope no relation's output carries (binding None -> left as-is).
+            # own output. Join conditions and the filters of a projected or
+            # emitting read may use a different row with no relation to anchor
+            # (binding None -> left as-is).
             single_input = _child_rel(*children[0]) if len(children) == 1 else None
             reducing = single_input is None and _is_reducing_join(node)
+            projected_read = rel_type == "read" and (
+                node.HasField("projection")
+                or node.common.WhichOneof("emit_kind") == "emit"
+            )
```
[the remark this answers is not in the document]
alexandrefimov: Right, done in cc69165.

src/substrait/type_inference.py
```diff
@@ -684,6 +684,56 @@ def _set_output_struct(op_name: str, inputs: list) -> stt.Type.Struct:
...
+            raise ValueError(
+                f"Read projection field index {item.field} is out of range "
+                f"for a struct with {len(struct.types)} fields"
+            )
+        field = struct.types[item.field]
+        if item.HasField("child"):
+            field = _project_read_type(field, item.child)
+        fields.append(field)
+    result = stt.Type.Struct()
+    result.CopyFrom(struct)
```
nielspardon: Drop `type_variation_reference` here, or correct the docstring on line 690. A variation describes the layout of the 3-field row, so carrying it onto a 1-field struct names a variation never declared for that type — substrait-java (`TypeCreator.of(structType.nullable()).struct(…)`) and substrait-validator (`data::new_struct(fields, root.nullable())`) both keep only nullability, so "matching Java and keeping struct metadata" is accurate for the ordering but not for the metadata. The emit tail already rebuilds the struct without it, so the same read reports variation 7 or 0 depending on whether an emit is present.

No suggestion block: `stt.Type.Struct(types=fields, nullability=struct.nullability)` is the one-line form, but it flips 7 assertions in `test_read_projection.py`, so the test expectations move with it.

src/substrait/type_inference.py
```diff
@@ -684,6 +684,56 @@ def _set_output_struct(op_name: str, inputs: list) -> stt.Type.Struct:
...
+    if field_kind != kind:
+        raise ValueError(
+            f"Read projection {kind} selection requires a {kind} type, got {field_kind}"
+        )
+    result = stt.Type()
+    result.CopyFrom(field)
+    if kind == "struct":
+        result.struct.CopyFrom(_project_read_struct(field.struct, select.struct))
+    else:
+        # List positions and map keys filter values without changing their type.
```
nielspardon: Scope this comment to what the spec actually leaves open. The mask section's Unwrapping Behavior says a single-element list selection removes the list by default and says nothing at all about map keys, so "without changing their type" reads as a rule where the spec has a TBD — the behavior itself matches both substrait-java and substrait-validator, so only the wording needs to say it's a compatibility choice rather than the spec's answer.

src/substrait/utils/__init__.py
```diff
@@ -604,15 +607,22 @@ def convert_rel(rel, scope):
...
+            projected_read = rel_type == "read" and (
+                node.projection.HasField("select")
+                or node.common.WhichOneof("emit_kind") == "emit"
+            )
             for name, expr in _iter_named_direct_expressions(node):
                 if single_input is not None:
                     binding = single_input
                 elif reducing and name in _JOIN_COMBINED_SCOPED_FIELDS:
                     binding = None
+                elif projected_read and name in ("filter", "best_effort_filter"):
```
nielspardon: Hoist this to a module-level `frozenset` beside `_JOIN_COMBINED_SCOPED_FIELDS`, which the line two above already uses for the same kind of rule.

src/substrait/type_inference.py
```diff
@@ -696,6 +746,13 @@ def infer_rel_schema(rel: stalg.Rel, *, registry=None, subtrees=()) -> stt.Type.
 
     if rel_type == "read":
         (common, struct) = (rel.read.common, rel.read.base_schema.struct)
+        if rel.read.projection.HasField("select"):
```
nielspardon: `src/substrait/utils/display.py:205` still resolves field names through the unprojected `read.base_schema.names`, so it now disagrees with this function: for a read of `[id, txt, flag]` masked to `[2]` with a filter referencing field 0, inference says `bool` (`flag`) while `pretty_print_plan` prints `field: id`. Before this PR both were projection-blind and agreed; a `names` counterpart via `type_num_names` would resolve it, in this PR or a follow-up.

nielspardon (commented):
Decide whether a pruned struct should keep its `type_variation_reference` — that's the one open question left; the rest below is wording and a stale sibling.

src/substrait/type_inference.py
alexandrefimov: Dropped in fdce125: a projected struct keeps only its nullability, and the seven expectations moved with it.

src/substrait/type_inference.py
alexandrefimov: Reworded in fdce125 as a compatibility choice.

src/substrait/utils/__init__.py
alexandrefimov: Done in fdce125.

src/substrait/type_inference.py
alexandrefimov: Fixed here in fdce125: the printer takes the selected fields' names in mask order. A child mask still leaves the nested names in a block as they are.

src/substrait/utils/display.py
```diff
@@ -5,10 +5,32 @@
...
+    selects whole blocks in its order; a child mask does not prune a block.
+    """
+    names = list(read.base_schema.names)
+    if not read.projection.HasField("select"):
+        return names
+    lengths = [type_num_names(t) for t in read.base_schema.struct.types]
+    starts = [0, *itertools.accumulate(lengths)]
+    return [
+        name
+        for item in read.projection.select.struct_items
```
nielspardon: Skip a mask item whose field is out of range. `starts[item.field + 1]` raises `IndexError` on an out-of-range index and a negative one silently returns an empty name list, where the printer previously tolerated any mask — and `_resolve_field_name` twenty lines up deliberately degrades to `field_N` rather than raising, which is the behavior you want from a tool people reach for when a plan is already malformed.

```suggestion
        for item in read.projection.select.struct_items
        if 0 <= item.field < len(lengths)
        for name in names[starts[item.field] : starts[item.field + 1]]
```

nielspardon (commented):
One guard left, then this looks good to me.

src/substrait/utils/display.py
alexandrefimov: Applied in 28450e8, with a test for out-of-range and negative mask items.

src/substrait/utils/display.py
```diff
@@ -5,10 +5,32 @@
...
+        for name in names[starts[item.field] : starts[item.field + 1]]
```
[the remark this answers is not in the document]
alexandrefimov: The printer ignored a read's emit before this PR too, so I'm leaving that for a separate change.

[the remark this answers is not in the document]
nielspardon: Worth one correction before this is parked: the pre-PR printer ignored projection *and* emit, and for some plans those two cancelled out, so honoring projection alone flips which reads are mislabelled rather than leaving them as they were.

| mask, emit | correct | `main` | this PR |
| --- | --- | --- | --- |
| `[2,0]`, `[1]` | `id` | `id` ✓ | `flag` ✗ |
| `[2,0]`, `[0]` | `flag` | `id` ✗ | `flag` ✓ |
| `[1,2]`, `[1]` | `flag` | `id` ✗ | `txt` ✗ |

Neither ref is right in general, so deferring is reasonable — just noting the first row regresses, in case that changes the call.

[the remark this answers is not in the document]
alexandrefimov: Right, deferring would regress your first row. ae7f299 applies the emit after the mask, with your table and an emit-only read as test cases.

src/substrait/type_inference.py
nielspardon: The nested behaviour matches substrait-java and the validator, but that is thinner ground than it looks: substrait-go round-trips the flag without acting on it, substrait-rs has no mask handling at all, and the field has never been modified since it was first added, with the spec's `[TBD how we express this in the serialization formats]` marker untouched since the docs move. The prose also defines two unwrapping rules and the proto opts out of only one — there is no list counterpart — while the `MaskExpression` comment in `algebra.proto` says a mask does not alter structure, which reads against unwrapping happening at all. So nobody has needed this yet rather than anyone having agreed on it; filed substrait-io/substrait#1230 to get a reading.

One thing that is decided: at the top level the validator parses a `ReadRel` projection with `struct_required = true` and rejects a singular mask with `maintain_singular_struct = false` (`TypeStructRequired`), which this infers a one-field struct for regardless. Not a blocker — keeping the struct is the only thing `infer_rel_schema` can return — but it means accepting a plan the reference validator refuses.

nielspardon (approved):
LGTM

PR #272, opened 2026-09-20
fix!: make a join's null-padded side nullable

### LINKED ISSUE #267: infer_rel_schema keeps outer and single join fields required on the null-padded side
`infer_rel_schema` preserves the input field nullabilities for LEFT, RIGHT, OUTER, LEFT_SINGLE and RIGHT_SINGLE joins. A required field on the null-padded side stays required, even though that join can return NULL for it.

Reproduced with substrait 0.31.0. `main` at 2f138c1 carries the same `type_inference.py`: the one commit between them touches a CI workflow.

The [Join Operation rules in spec v0.99.0](https://github.com/substrait-io/substrait/blob/v0.99.0/site/docs/relations/logical_relations.md#join-operation) describe the null padding for unmatched rows. It needs to be reflected in the output field types.

### Reproducer

```python
from substrait import algebra_pb2 as a, type_pb2 as t
from substrait.type_inference import infer_rel_schema

R, N = t.Type.NULLABILITY_REQUIRED, t.Type.NULLABILITY_NULLABLE

def read(name):
    return a.Rel(read=a.ReadRel(
        named_table=a.ReadRel.NamedTable(names=[name]),
        base_schema=t.NamedStruct(
            names=[name],
            struct=t.Type.Struct(
                types=[t.Type(i64=t.Type.I64(nullability=R))],
                nullability=R,
            ),
        ),
    ))

for kind in ("INNER", "LEFT", "RIGHT", "OUTER", "LEFT_SINGLE", "RIGHT_SINGLE"):
    rel = a.Rel(join=a.JoinRel(
        left=read("l"), right=read("r"),
        type=a.JoinRel.JoinType.Value("JOIN_TYPE_" + kind),
        expression=a.Expression(literal=a.Expression.Literal(boolean=True)),
    ))
    schema = infer_rel_schema(rel)
    print(kind, ["N" if f.i64.nullability == N else "R" for f in schema.types])
```

Every row prints `['R', 'R']`. The expected patterns are:

| Join | Expected |
| --- | --- |
| INNER | R R |
| LEFT / LEFT_SINGLE | R N |
| RIGHT / RIGHT_SINGLE | N R |
| OUTER | N N |

I also checked all four combinations of input nullability for these six join types: 11 of 24 schemas differ. INNER and already-nullable sides behave as expected, and inference does not mutate the inputs.

`_join_struct_from_schemas` currently concatenates the two input type lists for these `JoinRel` types without making the null-padded side nullable. The mark-join output columns are separate: #263 reports them and #265 changes them.

alexandrefimov:
`infer_rel_schema` keeps a join's fields as its inputs declare them, so a required field on the side the join fills with nulls stays required. Plans that DuckDB and Isthmus write for an ordinary `LEFT JOIN` hit this.

Make the padded side nullable, per the [Join Types in spec v0.99.0](https://github.com/substrait-io/substrait/blob/v0.99.0/site/docs/relations/logical_relations.md#join-types): `LEFT` and `LEFT_SINGLE` pad the right input, `RIGHT` and `RIGHT_SINGLE` the left, and `OUTER` both. Inner, semi, anti and mark joins pad nothing, a side that is already nullable is unchanged, and lateral joins share the same derivation.

Closes #267

BREAKING CHANGE: Outer and single joins now infer nullable fields on the null-padded side. A schema that repeated a required declaration there changes, along with anything derived from it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Join schema inference now accurately marks fields as nullable when unmatched rows can produce null values.
  * Left and right joins update nullability only on the opposite side; full outer joins update both sides.
  * Inner, semi, anti, and mark joins preserve existing field nullability.
  * Lateral left joins now correctly reflect nullable fields on the right side.
* **Tests**
  * Added coverage for join types and null-padding behavior.


src/substrait/type_inference.py
```diff
@@ -548,6 +548,33 @@ def _join_column_shape(type_name: str) -> str:
...
+        return ("left", "right")
+    return ()
+
+
+def _null_padded_types(struct: stt.Type.Struct, padded: bool) -> list:
+    """``struct``'s field types, made nullable when the join pads that side."""
+    if not padded:
+        return list(struct.types)
+    return [
+        _with_field_nullability(t, stt.Type.NULLABILITY_NULLABLE) for t in struct.types
```
nielspardon: Skip the `unbound` placeholder here — it is the one `Type.kind` with no `nullability` field, so a `LEFT`/`RIGHT`/`OUTER`/`*_SINGLE` join over a partially bound plan now raises `AttributeError: Protocol message Unbound has no "nullability" field` where it inferred fine before. Spec v0.99.0 `types/type_classes.md` allows `unbound` "as a field type within `ReadRel.base_schema`" and says it carries no nullability, so leaving it untouched is correct.

The fix belongs in `_with_field_nullability` (`src/substrait/type_inference.py:660`), outside this diff, so no suggestion block:

```python
    kind = out.WhichOneof("kind")
    if kind and kind != "unbound":
        getattr(out, kind).nullability = nullability
```

tests/test_type_inference.py
```diff
@@ -277,6 +277,63 @@ def test_inference_join_inner():
...
+            right=right_read_rel,
+            type=stalg.JoinRel.JOIN_TYPE_OUTER,
+            expression=None,
+        )
+    )
+
+    infer_rel_schema(rel)
+
+    assert infer_rel_schema(read_rel) == struct
+    assert infer_rel_schema(right_read_rel) == right_struct
```
nielspardon: Assert on the join's own inputs — `stalg.JoinRel(left=read_rel, ...)` deep-copies them, so this test cannot fail: an in-place `_with_field_nullability` (no `CopyFrom`) still passes it. The invariant is worth pinning, because `infer_rel_schema` returns a `ReadRel`'s `base_schema.struct` by reference.

```suggestion
    assert infer_rel_schema(rel.join.left) == struct
    assert infer_rel_schema(rel.join.right) == right_struct
```

src/substrait/type_inference.py
```diff
@@ -577,7 +604,10 @@ def _join_struct_from_schemas(
     elif shape in ("right", "right+mark"):
         types = list(right.types)
     else:
-        types = list(left.types) + list(right.types)
+        padded = _join_null_padded_sides(type_name)
+        types = _null_padded_types(left, "left" in padded) + _null_padded_types(
+            right, "right" in padded
```
nielspardon: Should a lateral join reach this at all for a right-oriented type? `_lateral_join_output_struct` shares this path, so a `LateralJoinRel` with `JOIN_TYPE_RIGHT`/`RIGHT_SINGLE`/`OUTER` now null-pads its left input — but spec v0.99.0 restricts lateral joins to `INNER`, `LEFT`, `LEFT_SEMI`, `LEFT_ANTI`, `LEFT_SINGLE`, `LEFT_MARK`, and a lateral join's left row always exists. `DataFrame.lateral_join` rejects those types, but `builders.plan.lateral_join` and raw protos do not, so inference silently invents nullability for a relation that cannot exist.

src/substrait/type_inference.py
```diff
@@ -548,6 +548,33 @@ def _join_column_shape(type_name: str) -> str:
...
+    if type_name in ("JOIN_TYPE_LEFT", "JOIN_TYPE_LEFT_SINGLE"):
+        return ("right",)
+    if type_name in ("JOIN_TYPE_RIGHT", "JOIN_TYPE_RIGHT_SINGLE"):
+        return ("left",)
+    if type_name == "JOIN_TYPE_OUTER":
+        return ("left", "right")
+    return ()
+
+
+def _null_padded_types(struct: stt.Type.Struct, padded: bool) -> list:
```
nielspardon: Consider one table keyed by join-type name instead of the helper pair — `{name: (pad_left, pad_right)}` read by both `join_output_names` and `_join_struct_from_schemas`. As written, the stringly-typed side markers are checked with `"left" in padded` at line 608, so a future edit returning `"both"` would silently disable all padding with no type checker in CI to catch it; `padded` here also shadows the `padded: bool` parameter two lines down, and `if not padded` re-decides what the caller already decided.

src/substrait/type_inference.py
```diff
@@ -548,6 +548,33 @@ def _join_column_shape(type_name: str) -> str:
     return "both"  # inner / outer / left / right / single
 
 
+def _join_null_padded_sides(type_name: str) -> tuple:
```
nielspardon: Three follow-ups that sit in unchanged regions, so no suggestions are attached:

- `src/substrait/type_inference.py:539` — "Single source of truth for the inferred type list and the RelRoot names" is no longer true now that a third name-keyed ladder must agree with the other two.
- `src/substrait/utils/__init__.py:473` — "The join relation's own output equals that combined row for every non-reducing join" is now false: for `JOIN_TYPE_LEFT` over two required `i64` columns the condition scope is `[REQUIRED, REQUIRED]` while the join's own output is `[REQUIRED, NULLABLE]`, so a correlated reference anchored to the join resolves against the padded output rather than the condition row it was written against.
- `tests/builders/plan/test_physical.py:61` — the only physical-join schema assertion projects to `WhichOneof("kind")`, which is blind to nullability, and the suite exercises no padding join type; the physical enums renumber the same names (`LEFT_SINGLE` is 7 in `JoinRel`, 9 in the other three), so the name-keyed dispatch this fix relies on is load-bearing and untested.

nielspardon (commented):
Widen the `BREAKING CHANGE:` footer to name left and right joins as well — plain `LEFT` and `RIGHT` change too, and they are the DuckDB/Isthmus case the PR opens with.

The padding table itself is right and matches substrait-java's `Join.deriveRecordType()` and substrait-go — though spec v0.99.0 states only the row-level padding, not the output *type* nullability, so that step is inferred rather than spec-defined.

src/substrait/type_inference.py
nielspardon: The same gap in `_field_nullability` is pre-existing (it makes any `SetRel` over an unbound column raise) and is now tracked as #276, so this PR only needs the `_with_field_nullability` guard.

alexandrefimov: Done in 7283813, with a test.

tests/test_type_inference.py
alexandrefimov: Applied in 7283813.

src/substrait/type_inference.py
alexandrefimov: No. 7283813 raises ValueError for them, as DataFrame.lateral_join does.

src/substrait/type_inference.py
alexandrefimov: Done in 7283813. join_output_names needs only the column shape, so it keeps _join_column_shape.

src/substrait/type_inference.py
alexandrefimov: All three in 7283813: docstring fixed, padding-join conditions stay offset-based like semi/anti/mark, and a physical-join test checks padding by type name.

nielspardon (commented):
All five addressed — threads resolved. The padding table now matches across all 13 join type names for `JoinRel` and the three physical joins, and the lateral join accepts exactly the six the spec allows.

